PR #67, opened 2022-03-11
Implement `$bitsAllClear`

w84thesun:


seeforschauer (commented):
would you add tests for int64 and for wrong values (i.e. string)?

w84thesun:
> would you add tests for int64 and for wrong values (i.e. string)?

Done

cmd/dance/main_unix.go
```diff
@@ -0,0 +1,29 @@
...
+package main
+
+import (
+	"context"
+	"os/signal"
+
+	"golang.org/x/sys/unix"
+)
+
+func notifyAppTermination(parent context.Context) (context.Context, context.CancelFunc) {
```
AlekSi, line 27: That should be in a separate PR

AlekSi (commented):
Tests fail

cmd/dance/main_unix.go
w84thesun: Ok, I'll move it to another PR.

w84thesun:
> > Done
> 
> Please re-request review in that case.
> 
> But also: tests fail.

Yup, I figure it out after I've added tests. I'm finishing https://github.com/FerretDB/FerretDB/issues/347 than will take a look at this.

AlekSi (commented):
There are conflicts now

tests/ferret.yml
```diff
@@ -5,7 +5,7 @@ results:
   common:
     stats:
       expected_skip: 1
-      expected_pass: 33
+      expected_pass: 3
```
AlekSi, line 8: That has to be a mistake

w84thesun: Yeah, there should be 38. Fixed it

PR #83, opened 2022-03-21
Projection: excluded and included fields

seeforschauer:


tests/ferret/core_test.go
```diff
@@ -54,7 +54,6 @@ func TestCore(t *testing.T) {
 			var writeErr mongo.BulkWriteException
 			assert.ErrorAs(t, err, &writeErr)
 			assert.True(t, writeErr.HasErrorCode(11000))
-
```
AlekSi, line 57: Let's keep that line, it is good :D

tests/ferret/core_test.go
```diff
@@ -326,10 +372,15 @@ func TestCore(t *testing.T) {
 				require.NotNil(t, cursor)
 
 				var expected []bson.D
+
 				for _, id := range tc.IDs {
 					v, ok := data[id]
 					require.True(t, ok)
-					expected = append(expected, bson.D{{"_id", id}, {"value", v}})
+					if tc.v != nil {
+						expected = append(expected, tc.v)
```
AlekSi, line 380: I think it makes it much more complicated than it is worth. Here we iterate over IDs, get the value by it – and then ignore it. Why?

And also why do we do that at all? All those values are available by IDs, that was the point – to avoid copying values.

seeforschauer: If to copy values only, then cannot check that the projection works.
in `TestCore/QueryOperators/ProjectionExclusion` the value is not included in the result

seeforschauer: logic fixed :)

tests/ferret/core_test.go
```diff
@@ -326,10 +369,14 @@ func TestCore(t *testing.T) {
 				require.NotNil(t, cursor)
 
 				var expected []bson.D
-				for _, id := range tc.IDs {
-					v, ok := data[id]
-					require.True(t, ok)
-					expected = append(expected, bson.D{{"_id", id}, {"value", v}})
+				if tc.v != nil {
+					expected = append(expected, tc.v)
```
AlekSi, line 373: Let's make an assignment (not `append`) to make it clearer

tests/ferret/core_test.go
```diff
@@ -153,6 +153,8 @@ func TestCore(t *testing.T) {
 		testCases := []struct {
 			name string // TODO move to map key
 			q    bson.D
+			o    *options.FindOptions
+			v    bson.D
```
AlekSi, line 157: I think it is worth mentioning what is this, why we need it in a comment.

And then I think we should check that only one of (v, IDs, err) is set (and mention that in the comment too)

tests/ferret/core_test.go
```diff
@@ -217,6 +219,41 @@ func TestCore(t *testing.T) {
...
+			{
+				name: "ProjectionExclusion",
+				q: bson.D{
+					{"_id", "double"},
+				},
+				o: options.Find().SetProjection(bson.D{{"value", false}}),
+				v: bson.D{
+					{"_id", "double"},
+				},
+				IDs: []string{"double"},
```
AlekSi, line 240: IDs are not used when v is set

tests/ferret/core_test.go
```diff
@@ -310,11 +365,17 @@ func TestCore(t *testing.T) {
 			t.Run(tc.name, func(t *testing.T) {
 				t.Parallel()
 
-				if (tc.IDs == nil) == (tc.err == nil) {
+				if (tc.IDs == nil) == (tc.err == nil) == (tc.v == nil) {
 					t.Fatal("exactly one of IDs or err must be set")
```
AlekSi, line 369: That error message is no longer accurate

tests/ferret/core_test.go
```diff
@@ -150,9 +150,15 @@ func TestCore(t *testing.T) {
...
 
+		// o is FindOptions, it's used in projections testing.
+		// v is a value and it is also being used in the projections testing,
+		// since the resulting value for projection differs from the document value.
+		// It's being checked only one of (v, IDs, err).
 		testCases := []struct {
 			name string // TODO move to map key
 			q    bson.D
+			o    *options.FindOptions
+			v    bson.D
```
AlekSi, line 161:
```suggestion
			v    bson.D // expected value; useful for testing projections
```

tests/ferret/core_test.go
```diff
@@ -150,9 +150,15 @@ func TestCore(t *testing.T) {
...
 		}
 
+		// o is FindOptions, it's used in projections testing.
+		// v is a value and it is also being used in the projections testing,
+		// since the resulting value for projection differs from the document value.
+		// It's being checked only one of (v, IDs, err).
 		testCases := []struct {
 			name string // TODO move to map key
 			q    bson.D
+			o    *options.FindOptions
```
AlekSi, line 160:
```suggestion
			o    *options.FindOptions // if empty, defaults to sorting by value for stable tests
```

tests/ferret/core_test.go
AlekSi, line 161: Should it be []bson.D for generality?

tests/ferret/core_test.go
```diff
@@ -217,6 +223,55 @@ func TestCore(t *testing.T) {
...
+				q: bson.D{
+					{"_id", "document-diverse"},
+				},
+				o: options.Find().SetProjection(bson.D{
+					{"document_id", false},
+					{"array", true},
+				}),
+				err: mongo.CommandError{
+					Code:    31253,
+					Name:    "ErrorCode(31253)",
```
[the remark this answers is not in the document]
AlekSi: https://github.com/FerretDB/FerretDB/pull/384#discussion_r832154632

tests/ferret/core_test.go
```diff
@@ -150,10 +150,16 @@ func TestCore(t *testing.T) {
 			require.NoError(t, err)
 		}
 
+		// o is FindOptions, it's used in projections testing.
+		// v is a value and it is also being used in the projections testing,
+		// since the resulting value for projection differs from the document value.
+		// It's being checked only one of (v, IDs, err).
```
AlekSi, line 156: And with inline comments we no longer need that comment

AlekSi (commented):
Tests fail

AlekSi:
https://github.com/FerretDB/dance/runs/5646312990?check_suite_focus=true#step:7:184

tests/ferret/core_test.go
seeforschauer: removed

seeforschauer:
hmmm, it didn't fail locally :|

AlekSi (approved):

PR #84, opened 2022-03-22
Add test for sort aggregation

w84thesun:


w84thesun:
I've checked errors. It seems that they are not related to changed code.

tests/ferret/aggregation_test.go
```diff
@@ -0,0 +1,124 @@
...
+				name: "SortBoroughsPassFloat64Value",
+				sort: bson.D{{"borough", float64(1.0)}, {"name", int64(1)}},
+				IDs:  []string{"3", "5", "1", "4", "2"},
+			},
+		}
+
+		for _, tc := range testCases {
+			tc := tc
+
+			if tc.name == "" {
```
AlekSi, line 81: I think we better do `require.NotEmpty(t, tc.name)` to avoid mutating test cases

tests/ferret/aggregation_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+				require.NoError(t, err)
+				require.NotNil(t, cursor)
+
+				var actual []bson.D
+				require.NoError(t, cursor.All(ctx, &actual))
+				var actualIDs []string
+				for _, d := range actual {
+					for k, v := range d.Map() {
+						if k == "_id" {
```
seeforschauer, line 105: here you can remove complexity

tests/ferret/aggregation_test.go
```diff
@@ -0,0 +1,121 @@
...
+			err  error
+		}{
+			{
+				name: "SortBoroughs",
+				sort: bson.D{{"borough", int32(1)}},
+				IDs:  []string{"3", "5", "1", "4", "2"},
+			},
+			{
+				name: "SortBoroughsAndNames",
+				sort: bson.D{{"borough", int32(1)}, {"name", int32(1)}},
```
seeforschauer, line 62: what the behavior will be if put negative value? 
what if put a string value? error?

seeforschauer (changes requested):
there are conflicts to resolve.

tests/ferret/aggregation_test.go
w84thesun: fixed

seeforschauer (dismissed):

AlekSi (changes requested):
Tests fail

AlekSi:
I will close that one as we will add proper tests at #423.

Issue #134, opened 2022-06-20
Update Mongo Go driver tests configuration


AlekSi:
* Get familiar with this tool.
* Update `tests/mongo-go-driver` submodule.
* Update tests configuration.
* Try to enable more packages.
* Think about what could be improved.

Spend no more than one day.

AlekSi:
@noisersup Correct

noisersup:
Is there anything I should have in mind while deciding which package I implement (like time complexity and more)?

For an example, after enabling `/mongo/integration` the number of tests jumped from **227** to **1390** and the runtime from **0m2.249s** to **1m50.866s**

AlekSi:
I don't have any hidden knowledge to share on that topic. That's where the last point from the description becomes the most important :)

We sure want to run more driver's integration tests against FerretDB to track our progress. But that (and updating configs) should be less painful.

PR #137, opened 2022-06-23
Update Mongo Go driver tests configuration

noisersup:
Refs #134.

### DIFF
--- a/tests/mongo-go-driver.yml
+++ b/tests/mongo-go-driver.yml
@@ -136,7 +136,8 @@ results:
       # skip if no compressor specified
       - go.mongodb.org/mongo-driver/x/mongo/driver/integration/TestCompression
 
+      # always skipped
+      - go.mongodb.org/mongo-driver/x/mongo/driver/integration/TestInsert
+
       # skip because authentication is required
       - go.mongodb.org/mongo-driver/x/mongo/driver/integration/TestSCRAM
-
-      - go.mongodb.org/mongo-driver/x/mongo/driver/integration/TestInsert


tests/mongo-go-driver.yml
```diff
@@ -67,14 +71,35 @@ results:
...
+      # skip if CRYPT_SHARED_LIB_PATH is not set
+      - go.mongodb.org/mongo-driver/mongo/TestClient/mongocryptd_or_crypt_shared
+
+      # skip if no compressor specified
+      - go.mongodb.org/mongo-driver/x/mongo/driver/integration/TestCompression
+
+      # skip because authentication is required
+      - go.mongodb.org/mongo-driver/x/mongo/driver/integration/TestSCRAM
+
+      - go.mongodb.org/mongo-driver/x/mongo/driver/integration/TestInsert
```
noisersup, line 93: I'm not sure how to comment that line.
The test just [skips on the first line of it](https://github.com/mongodb/mongo-go-driver/blob/master/x/mongo/driver/integration/insert_test.go#L20)

AlekSi: I think that could be something like "Always skipped"

AlekSi (commented):
Looks good. Please add those comments about skipped tests

AlekSi:
Any ideas how to improve that part shared between configurations?

noisersup:
In that topic I was thinking about a few approaches:
- Move the `skip` section to seperate block and adjust a code to parse it in both databases. In fact I suppose that this [common field](https://github.com/FerretDB/dance/blob/main/internal/config.go#L39) could be an attempt to do that (correct me if I'm wrong)
- A yaml generation (something like jsonnet where we can just declare a variable and "compile" to yaml). I can make a little research on that if needed.

noisersup:
Also if we are talking about improvements. Is there anything that stops us from automating submodule updates?

AlekSi:
Dependabot already does that: https://github.com/FerretDB/dance/blob/edf7f8560a91eeadabf39b25b189c13aa392ce88/.github/dependabot.yml#L25-L31

AlekSi:
See #139, for example

AlekSi:
See #142

AlekSi (approved):

PR #164, opened 2022-07-06
Validate collection names

seeforschauer:
Closes FerretDB/FerretDB#806.

| Case | FerretDB | MongoDB | Status | 
|--|--|--|--| 
| `system.` |  fails: `system.` contains a dot |  fails: `system.` is a reserved prefix | skipped as both fail | 
| `_ferretdb_` | fails: `_ferretdb_` is a reserved prefix | pass | added | 
|  max length | 120  | 255 |  added | 
|  dots | fail | pass |  added | 
|  dashes | fail | pass |  added | 

Corresponding FerretDB PR [#844](https://github.com/FerretDB/FerretDB/pull/844).

### DIFF
--- a/tests/diff/collection_name_test.go
+++ b/tests/diff/collection_name_test.go
@@ -27,7 +27,8 @@ import (
 //   * dots
 //   * dashes
 //   * max length in FerretDB: 255, in MongoDB: 63.
-//   * FerretDB reserved prefix is _ferretdb_, MongoDB - system..
+//   * FerretDB reserved prefix is _ferretdb_, so FerretDB doesn't allow such prefixes.
+//  MongoDB's reserved prefix is 'system.'. However FerretDB doesn't allow such name because of a dot in it.
 func TestCollectionName(t *testing.T) {
 	t.Parallel()
 
--- a/tests/diff/collection_name_test.go
+++ b/tests/diff/collection_name_test.go
@@ -44,8 +44,6 @@ func TestCollectionName(t *testing.T) {
 			ctx, db := setup(t)
 			err := db.CreateCollection(ctx, collection)
 			require.NoError(t, err)
-			err = db.Collection(collection).Drop(ctx)
-			require.NoError(t, err)
 		})
 	})
 
@@ -66,8 +64,6 @@ func TestCollectionName(t *testing.T) {
 		t.Run("MongoDB", func(t *testing.T) {
 			ctx, db := setup(t)
 			err := db.CreateCollection(ctx, collection)
-			require.NoError(t, err)
-			err = db.Collection(collection).Drop(ctx)
 			require.NoError(t, err)
 		})
 	})
--- a/tests/diff/collection_name_test.go
+++ b/tests/diff/collection_name_test.go
@@ -39,6 +39,7 @@ func TestCollectionName(t *testing.T) {
 		collection := strings.Repeat("a", 200)
 
 		t.Run("FerretDB", func(t *testing.T) {
+			_ = db.Collection(collection).Drop(ctx)
 			err := db.CreateCollection(ctx, collection)
 			expected := mongo.CommandError{
 				Name:    "InvalidNamespace",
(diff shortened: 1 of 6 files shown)


rumyantseva (commented):
In general, looks good to me (added a tiny typo fix and a question).

tests/diff/collection_name_test.go
```diff
@@ -0,0 +1,126 @@
...
+// TestCollectionName documents difference in reposnes:
+//   * dots
+//   * dashes
+//   * max length in FerretDB: 255, in MongoDB: 63.
+//   * FerretDB reserved prefix is _ferretdb_, MongoDB - system..
+func TestCollectionName(t *testing.T) {
+	t.Parallel()
+
+	ctx, db := setup(t)
+	db = db.Client().Database("admin")
```
AlekSi, line 35: There is no need to use `admin` database, the per-test database returned by `setup` will do just fine. That also removes the need to drop collections

tests/diff/collection_name_test.go
```diff
@@ -0,0 +1,126 @@
...
+	dbName := db.Name()
+
+	t.Run("Dot", func(t *testing.T) {
```
AlekSi, line 38: We really need only one of those tests, just like we have only one test for different error messages

tests/diff/collection_name_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+// TestCollectionName documents difference in reposnes:
+//   * dots
+//   * dashes
+//   * max length in FerretDB: 255, in MongoDB: 63.
+//   * FerretDB reserved prefix is _ferretdb_, MongoDB - system..
```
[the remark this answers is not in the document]
seeforschauer: > I see in the description you decided to skip the test with the `system.` prefix. Should we mention `system` at all then?

I've reformulated it. 
It is worth mentioning `system.` bcz we are talking here about reserved words in FerretDB and in MongoDB.

tests/diff/collection_name_test.go
seeforschauer: In the case of a regex, we don't even check all filtered cases. I tried to cover the most often cases. 
But OK..

tests/diff/collection_name_test.go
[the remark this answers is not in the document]
seeforschauer: > > I see in the description you decided to skip the test with the `system.` prefix. Should we mention `system` at all then?
> 
> I've reformulated it. It is worth mentioning `system.` bcz we are talking here about reserved words in FerretDB and in MongoDB.

Removed the comment. You know better is an enemy of good.

tests/diff/collection_name_test.go
seeforschauer: > There is no need to use `admin` database.

Applied, looks good.

> That also removes the need to drop collections

However, It doesn't work w/o deletion.

rumyantseva (approved):
LGTM

tests/diff/collection_name_test.go
AlekSi: `setup` creates unique empty database for each test. There are no collections to drop.

If it does not work, please demonstrate that

seeforschauer: > `setup` creates unique empty database for each test. There are no collections to drop.
> 
> If it does not work, please demonstrate that

Well, it was not defined clearly what you asked about, so I tried if it work with a database created for 2 tests, not for each separately. 

If you want to prove it, you can run dance tests against MongoDB with setup placed under `t.Run("ReservedPrefix", func(t *testing.T) {` not under `t.Run("FerretDB", func(t *testing.T) {`. 


Of course, when we initialize a database each time for each test, it works now.

tests/diff/collection_name_test.go
```diff
@@ -0,0 +1,74 @@
...
+				Message: fmt.Sprintf(`Invalid collection name: '%s.%s'`, dbName, collection),
+			}
+			AssertEqualError(t, expected, err)
+		})
+
+		t.Run("MongoDB", func(t *testing.T) {
+			ctx, db := setup(t)
+			err := db.CreateCollection(ctx, collection)
+			require.NoError(t, err)
+			err = db.Collection(collection).Drop(ctx)
```
AlekSi, line 47: There is no need to drop collection – the whole database will be dropped after the test

seeforschauer: > There is no need to drop collection – the whole database will be dropped after the test

removed.

AlekSi (approved):

PR #153, opened 2022-06-30
Enable Go driver's integration tests

noisersup:
Refs #144.

PR #170, opened 2022-07-11
Add regex support for test names

### LINKED ISSUE #169: Add regex support for test names
https://github.com/FerretDB/dance/blob/main/tests/diff.yml

If there is a support for regular expressions in test names, that file could be simplified. For example, we could use the following syntax:
```
pass:
  - regex: FerretDB$
```

noisersup:
Closes #169.

- [x] `output_regex` support
- [x] correct duplication handling.

### DIFF
--- a/internal/tests_config.go
+++ b/internal/tests_config.go
@@ -178,7 +178,15 @@ func (tc *TestsConfig) getExpectedStatusRegex(testName string, result *TestResul
 				continue
 			}
 			if matchedRegex != "" {
-				panic(fmt.Sprintf("test %s\n(output: %s)\nmatch more than one regexps: %s, %s", testName, result.Output, matchedRegex, reg))
+				panic(
+					fmt.Sprintf(
+						"test %s\n(output: %s)\nmatch more than one regexps: %s, %s",
+						testName,
+						result.Output,
+						matchedRegex,
+						reg,
+					),
+				)
 			}
 			matchedStatus = expectedRes.expectedStatus
 			matchedRegex = reg
@@ -191,7 +199,15 @@ func (tc *TestsConfig) getExpectedStatusRegex(testName string, result *TestResul
 				continue
 			}
 			if matchedRegex != "" {
-				panic(fmt.Sprintf("test %s\n(output: %s)\nmatch more than one regexps: %s, %s", testName, result.Output, matchedRegex, reg))
+				panic(
+					fmt.Sprintf(
+						"test %s\n(output: %s)\nmatch more than one regexps: %s, %s",
+						testName,
+						result.Output,
+						matchedRegex,
+						reg,
+					),
+				)
 			}
 			matchedStatus = expectedRes.expectedStatus
 			matchedRegex = reg


noisersup:
whoops, not yet

internal/yaml.go
```diff
@@ -0,0 +1,157 @@
...
+// It differs from it by using "any" type to be able to parse maps (i.e. "- output_regex: ...").
+//
+// To gain a data the struct should be first converted to TestsConfig with testsConfig.Convert() function.
+type testsConfig struct { // testsConfig -> TestsConfig
+	Default status `yaml:"default"`
+	Stats   *Stats `yaml:"stats"`
+	Pass    []any  `yaml:"pass"`
+	Skip    []any  `yaml:"skip"`
+	Fail    []any  `yaml:"fail"`
+}
```
noisersup, line 66: PR ready for review.
As you can see here we have 3 structs which are only used to fetch yaml data. Every of them has it's "internal" representation.
I can see here a little name inconsistency. We've discussed it with @seeforschauer and my point is that:
- The `Config` yaml representation adds a "File" part at the end.
- The `resultList` and `testsConfig` representation is just exported (which maybe would not be a bad idea if the structs were in the seperate package, but they're not).

I think that current state is a little bit misleading (especially for somone new to the codebase).

@AlekSi please let me know what do you think and maybe if you have some suggestions in that case.

seeforschauer: Yep, here something can be done better probably.

My point is that because we made an analogy for corresponded structures - it's easy to understand.
```go
// resultList -> ResultList
// testsConfig -> TestsConfig
```

It seems too little code to extract it into a separate package and make it unexported. 
And also probably will overcomplicate things.

AlekSi: I think having types `resultList` and `ResultList` that differ only in case is confusing, especially given that both of them are not used outside of the package. Suffix would be better, in my opinion. For example, `results` (that's not even a list) and `resultsYAML`.

AlekSi: Alternatively, `Results` and `ResultsYAML`. Both are exported, but they also were exported before this PR, so we just don't change it now. Maybe letter we will split them into a few packages

seeforschauer: > Alternatively, `Results` and `ResultsYAML`. Both are exported, but they also were exported before this PR, so we just don't change it now. Maybe letter we will split them into a few packages

Vote for this.

noisersup: Okay, but should we use "yaml" or maybe something less language specific like "file"?

AlekSi: Well, that's not a whole file, only part of it. It is also has only `yaml` tags, not `json` or `toml` tags, and we don't support anything other than YAML. So I propose to ~[paint this bikeshed green](https://bikeshed.org)~ call it `ResultsYAML`.

internal/yaml.go
```diff
@@ -0,0 +1,157 @@
...
+		Results{common, ferretDB, mongoDB},
+	}, nil
+}
+
+// Convert converts a TestsConfigYAML to the TestsConfig struct.
+// TestsConfigYAML is a yaml file with the tests.
+// TestsConfig is an internal representation of yaml test file.
+func (ftc *TestsConfigYAML) Convert() (*TestsConfig, error) {
+	if ftc == nil {
+		return nil, nil // not sure if that works
```
seeforschauer, line 96: > not sure if that works

what did you mean?

internal/yaml.go
```diff
@@ -0,0 +1,157 @@
...
+	return &Config{
+		cf.Runner,
+		cf.Dir,
+		cf.Args,
+		Results{common, ferretDB, mongoDB},
+	}, nil
+}
+
+// Convert converts a TestsConfigYAML to the TestsConfig struct.
+// TestsConfigYAML is a yaml file with the tests.
```
seeforschauer, line 92: > TestsConfigYAML is a yaml file with the tests

Hmmm, it's not a file it's a struct. No?

internal/yaml.go
```diff
@@ -0,0 +1,157 @@
...
+	if ftc == nil {
+		return nil, nil // not sure if that works
+	}
+
+	tc := TestsConfig{ftc.Default, ftc.Stats, Tests{}, Tests{}, Tests{}}
+
+	//nolint:govet // we don't care about alignment there
+	for _, testCategory := range []struct { // testCategory examples: pass, skip sections in the yaml file
+		yamlTests []any  // taken from the file, yaml representation of tests, incoming tests
+		outTests  *Tests // output tests
```
seeforschauer, line 104:
```suggestion
		outTests  *Tests // yamlTests transformed to the internal representation
```

internal/tests_config.go
```diff
@@ -71,30 +40,9 @@ type TestsConfig struct {
...
-				return nil, fmt.Errorf("duplicate test or prefix: %q", t)
-			}
-			res[t] = tcat.testsStatus
-		}
-	}
-
-	return res, nil
+	Names              []string // names (i.e. "go.mongodb.org/mongo-driver/mongo/...")
+	NameRegexPattern   []string // regex: "mongo.org/.*", the regex for the test name.
+	OutputRegexPattern []string // output_regex: "^server version \"5.0.9\" is (lower|higher).*".
```
seeforschauer, line 45: Is it a regex for test output? 
Then why we should fail if both `NameRegexPattern` and `OutputRegexPattern` match?

internal/tests_config.go
```diff
@@ -71,30 +40,9 @@ type TestsConfig struct {
 
 // Tests are the tests from yaml category pass / fail / skip.
```
seeforschauer, line 41: Let's add here rules: if  both `NameRegexPattern` and `OutputRegexPattern` match, parsing file fails.

internal/yaml.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// ConfigYAML is a yaml tests representation of the Config struct.
+//
+// It is used only to fetch data from file. To get any of
+// the dance configuration data it should be converted to
+// Config struct with Convert() function.
+//
+//nolint:govet // we don't care about alignment there
+type ConfigYAML struct { // ConfigYAML -> Config
```
seeforschauer, line 42: > // ConfigYAML -> Config
> // ResultsYAML -> Results
> // TestsConfigYAML -> TestsConfig

Let's move this to one place - the `Config` struct comment.
And let's turn it to the sentence, not a sentence fragment.

internal/yaml.go
```diff
@@ -0,0 +1,157 @@
...
+		cf.Runner,
+		cf.Dir,
+		cf.Args,
+		Results{common, ferretDB, mongoDB},
+	}, nil
+}
+
+// Convert converts a TestsConfigYAML to the TestsConfig struct.
+// TestsConfigYAML is a yaml file with the tests.
+// TestsConfig is an internal representation of yaml test file.
```
seeforschauer, line 93: > // TestsConfig is an internal representation of yaml test file.

The comment is for `TestsConfigYAML`. Let's start the comment with `TestsConfigYAML` word or with `Convert` word. Otherwise, I wonder if this is a comment for this function or if I was just copied from somewhere). 


```suggestion
// Convert validates yaml and converts TestsConfigYAML to the internal representation of yaml test file, TestsConfig.
```

seeforschauer:
@noisersup dance tests fail.

internal/tests_config.go
noisersup: https://github.com/FerretDB/dance/issues/169#issuecomment-1180519209

internal/yaml.go
noisersup: I left that comment on early stage of the PR. there were some problems which occured in tests.
The if statement just returns nil TestsConfig when the TestsConfigYAML is also a nil so I will remove the comment.

internal/tests_config.go
```diff
@@ -230,3 +155,88 @@ func (tc *TestsConfig) Compare(results *TestResults) (*CompareResult, error) {
 
 	return compareResult, nil
 }
+
+// getExpectedStatusRegex compiles result output with expected outputs and return expected status.
+// If no output matches expected - returns nil.
+// If both of the regexps match, returns the error.
```
seeforschauer, line 161: hm, it doesn't return the error: it panics.

internal/tests_config.go
```diff
@@ -230,3 +155,88 @@ func (tc *TestsConfig) Compare(results *TestResults) (*CompareResult, error) {
...
+
+	if t := strings.TrimRight(path, "/"); t != path {
+		return t
+	}
+
+	i := strings.LastIndexAny(path, "/.")
+	return path[:i+1]
+}
+
+func (tc *TestsConfig) toMap() (map[string]status, error) {
```
seeforschauer, line 222: Let's add a comment for this function also.

internal/yaml.go
```diff
@@ -0,0 +1,162 @@
...
+	ExpectedFail   int `yaml:"expected_fail"`
+	ExpectedSkip   int `yaml:"expected_skip"`
+	ExpectedPass   int `yaml:"expected_pass"`
+}
+
+// Structs and their internal equivalents:
+// - ConfigYAML -> Config
+// - ResultsYAML -> Results
+// - TestsConfigYAML -> TestsConfig.
+
```
seeforschauer, line 39: This empty line will divide the doc into 2 separate parts, do we really want this? 
Let's maybe add a `//` here.
```suggestion
//
```

noisersup: I think they should be seperated. When we add `//` this comment will be added specifically to the `ConfigYAML` struct (and not to `ResultYAML` and `TestsConfigYAML`).

noisersup:
I've made a fix to the [bug related to this PR](https://github.com/FerretDB/dance/pull/153#issuecomment-1186347680)

internal/tests_config.go
```diff
@@ -178,7 +178,15 @@ func (tc *TestsConfig) getExpectedStatusRegex(testName string, result *TestResul
...
-				panic(fmt.Sprintf("test %s\n(output: %s)\nmatch more than one regexps: %s, %s", testName, result.Output, matchedRegex, reg))
+				panic(
+					fmt.Sprintf(
+						"test %s\n(output: %s)\nmatch more than one regexps: %s, %s",
+						testName,
+						result.Output,
+						matchedRegex,
+						reg,
+					),
+				)
```
noisersup, line 189: @seeforschauer please let me know if that meets our code style requirements.

seeforschauer (approved):

rumyantseva (approved):
As usual, I'm not a big expert in dance, but the changes look reasonable to me.

PR #176, opened 2022-07-19
Validate database names

seeforschauer:
Closes https://github.com/FerretDB/FerretDB/issues/807.

Refs https://github.com/FerretDB/FerretDB/pull/913.

### DIFF
--- a/tests/diff.yml
+++ b/tests/diff.yml
@@ -11,7 +11,7 @@ results:
     pass:
       - github.com/FerretDB/dance/tests/diff/TestCollectionName/Length200/FerretDB
       - github.com/FerretDB/dance/tests/diff/TestCollectionName/ReservedPrefix/FerretDB
-      - github.com/FerretDB/dance/tests/diff/TestDataBaseName/ReservedPrefix/FerretDB
+      - github.com/FerretDB/dance/tests/diff/TestDatabaseName/ReservedPrefix/FerretDB
       - github.com/FerretDB/dance/tests/diff/TestErrorMessages/FerretDB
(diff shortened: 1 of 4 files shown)


tests/diff.yml
```diff
@@ -5,12 +5,14 @@ args: ["-timeout=20s", "-shuffle=on", "-race", "./..."]
 results:
   ferretdb:
     stats:
-      expected_fail: 9
-      expected_pass: 4
+      expected_fail: 14
+      expected_pass: 6
 
     pass:
       - github.com/FerretDB/dance/tests/diff/TestCollectionName/Length200/FerretDB
```
AlekSi, line 12: @noisersup can this be replaced with something like
```
- `regex`: /dance/tests/diff/.+/FerretDB$
```
?

noisersup: It should work let me try

seeforschauer: > @noisersup can this be replaced with something like
> 
> ```
> - regex: /dance/tests/diff/.+/FerretDB$
> ```
> 
> ?

it can be done in a separate PR.

AlekSi: It can be, but let's do that in this PR if it works – that will make diff configuration more clean

noisersup: I will do that in #177 then :)

seeforschauer: > It can be, but let's do that in this PR if it works – that will make diff configuration more clean

Test failed when @noisersup pushed.

w84thesun (dismissed):

w84thesun (approved):

AlekSi (approved):

PR #211, opened 2022-10-12
Add simplest validation to check data documents

rumyantseva:
This PR closes https://github.com/FerretDB/FerretDB/issues/693 and provides diff tests for https://github.com/FerretDB/FerretDB/pull/1246/files.

### DIFF
--- a/tests/diff.yml
+++ b/tests/diff.yml
@@ -5,8 +5,8 @@ args: ["-timeout=20s", "-shuffle=on", "-race", "./..."]
 results:
   ferretdb:
     stats:
-      expected_fail: 12
-      expected_pass: 5
+      expected_fail: 17
+      expected_pass: 7
 
     pass:
       - regex: FerretDB$
--- a/tests/diff.yml
+++ b/tests/diff.yml
@@ -17,7 +17,7 @@ results:
   mongodb:
     stats:
       expected_fail: 12
-      expected_pass: 5
+      expected_pass: 7
 
     pass:
       - regex: MongoDB$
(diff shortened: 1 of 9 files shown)


tests/diff/document_validation_test.go
```diff
@@ -0,0 +1,103 @@
...
+	collection := db.Collection("document-validation")
+	_, err := collection.InsertOne(ctx, bson.D{
+		{"_id", "valid"},
+		{"v", int32(42)},
+	})
+	require.NoError(t, err)
+
+	for name, tc := range map[string]struct {
+		expectedErrMongoDB  *mongo.CommandError
+		expectedErrFerretDB *mongo.CommandError
```
AlekSi, line 40: Notice how https://github.com/FerretDB/dance/blob/main/tests/diff/names_test.go and other tests don't have different expected errors. That's because of our test configuration at https://github.com/FerretDB/dance/blob/main/tests/diff.yml. See also https://github.com/FerretDB/dance/blob/main/CONTRIBUTING.md#running-tests

AlekSi (changes requested):

noisersup (commented):
Looks like it's not ready

chilagrow (approved):
Looks good to me!

tests/diff.yml
```diff
@@ -5,8 +5,8 @@ args: ["-timeout=20s", "-shuffle=on", "-race", "./..."]
 results:
   ferretdb:
     stats:
-      expected_fail: 12
-      expected_pass: 5
+      expected_fail: 14
+      expected_pass: 7
```
rumyantseva, line 9: @AlekSi we have one test for insert and one test for update, so we expect two more tests to fail, and two more tests to pass, right? Somehow, dance still fails, and I don't understand why.

rumyantseva: image

AlekSi (approved):

w84thesun (approved):

PR #219, opened 2022-10-26
Improve `names_tests.go`

noisersup:


tests/diff/names_test.go
```diff
@@ -43,9 +44,27 @@ func TestDatabaseName(t *testing.T) {
 		})
 
 		t.Run("MongoDB", func(t *testing.T) {
-			ctx, db := setup(t)
+			require.NoError(t, err)
```
chilagrow, line 47: For the readability, I prefer to have the code that is producing in the same `t.Run`. Just my preference :hugs:

AlekSi: But `err` is different in "FerretDB" subtest

AlekSi (approved):

PR #226, opened 2022-10-31
Make users know about telemetry via `startupWarnings`

rumyantseva:
Tests for https://github.com/FerretDB/FerretDB/pull/1336.
Issue: https://github.com/FerretDB/FerretDB/issues/1328.

### DIFF
--- a/tests/diff/startup_warnings_test.go
+++ b/tests/diff/startup_warnings_test.go
@@ -15,10 +15,11 @@
 package diff
 
 import (
+	"testing"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/bson"
-	"testing"
 )
 
 func TestStartupWarnigns(t *testing.T) {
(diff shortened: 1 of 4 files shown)


tests/diff/startup_warnings_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+func TestStartupWarnigns(t *testing.T) {
+	t.Parallel()
+
+	// TODO How to enable and disable telemetry when FerretDB is running?
```
rumyantseva, line 28: @AlekSi, and now I need some help understanding how to write dance diff tests for telemetry, as there is no way to set telemetry as true/false/undecided through db commands. So, what's the way to ensure that the telemetry state is set as we expect it?

AlekSi: The current way is "wait for FerretDB/FerretDB#1323", I guess :)

AlekSi (commented):
Please don't update mongo-go-driver in that PR

rumyantseva:
@noisersup I suggest closing this PR for now as it focuses on `undecided`, and we currently don't have a decision about a good approach to test it.

PR #252, opened 2022-11-23
Deprecate float values

w84thesun:
After [#1401](https://github.com/FerretDB/FerretDB/pull/1401)

Refs [#1151](https://github.com/FerretDB/FerretDB/issues/1151).

### DIFF
--- a/tests/diff/float_values_test.go
+++ b/tests/diff/float_values_test.go
@@ -1,0 +1,75 @@
+// Copyright 2021 FerretDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package diff
+
+import (
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"math"
+	"testing"
+)
+
+func TestFloatValues(t *testing.T) {
+	t.Parallel()
+
+	ctx, db := setup(t)
+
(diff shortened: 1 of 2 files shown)


chilagrow:
Tests are failing for FerretDB :thinking:

tests/diff/float_values_test.go
```diff
@@ -0,0 +1,76 @@
...
+		for name, tc := range map[string]struct {
+			doc      bson.D
+			expected mongo.CommandError
+		}{
+			"NaN": {
+				doc: bson.D{{"_id", "1"}, {"foo", math.NaN()}},
+				expected: mongo.CommandError{
+					Code:    2,
+					Name:    "BadValue",
+					Message: `wire.OpMsg.Document: validation failed for { _id: "1", foo: nan.0 } with: NaN is not supported`,
```
AlekSi, line 43: `nan.0`?

w84thesun: This output is from [FormatAnyValue](https://github.com/FerretDB/FerretDB/blob/d3cbadfa98be1c035dd318dafa9e5525f2aa9203/internal/types/path.go#L237). This is how MongoDB format it.

chilagrow (approved):

tests/diff/float_values_test.go
rumyantseva: That's an interesting question: on the one hand, as we don't support `NaN`s anymore, we don't need to support that formatting, and we can offer something better than `nan.0` for `NaN`s. On the other hand, someone who has experience with MongoDB might expect to see `nan.0` in the response 🤔

tests/diff/float_values_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+func TestFloatValues(t *testing.T) {
+	t.Parallel()
+
+	ctx, db := setup(t)
+
+	t.Run("Insert", func(t *testing.T) {
```
rumyantseva, line 31: This subtest is not necessary at the moment but might make sense in the future if decide to test not only inserts. So if you added in on purpose, let's keep it (and if you added it by mistake let's remove it).

rumyantseva (approved):
A couple of non-critical comments, and looks good to me otherwise.

noisersup (approved):
LGTM! Although we are not sure if it passes without merging FerretDB/FerretDB#1401

PR #255, opened 2022-11-24
Init tests for `mongodump` and `mongorestore` utility

### LINKED ISSUE #149: Add tests for `mongorestore` and `mongodump`
We should run `mongorestore` and `mongodump` with various flags to ensure we stay compatible with FerretDB. Let's add a new set of tests that would run binaries, not use their tests.

* https://www.mongodb.com/docs/database-tools/mongorestore/
* https://www.mongodb.com/docs/database-tools/mongodump/

noisersup:
Closes #149.


AlekSi:
@noisersup those are not diff tests, that should a separate directory

AlekSi (commented):
```
    extra_hosts:
      - "host.docker.internal:host-gateway"
```

And let's maybe switch from old `docker-compose` to new `docker compose`

AlekSi (commented):
Please remove dependency on github.com/docker/docker package. `docker-compose exec` or `docker compose exec` should be executed using `os.Exec`

noisersup:
@AlekSi Based on your changes, is running `mongodump` on `mongosh` container is still relevant idea?

AlekSi:
We already have a container with those tools, we don't need to create another one

noisersup:
Which container?

noisersup:
Current state: Locally on a main branch I get the following error:
> Failed: error dumping metadata: error getting indexes for collection `test.values`: invalid DocumentSequence: a non-document value was found in sequence

Though, I cannot replicate it on dance tests

CONTRIBUTING.md
```diff
@@ -1,7 +1,6 @@
 # Contributing
 
 The dance tool and tests run on the host; macOS, Linux, and Windows are expected to work.
-Databases under test (FerretDB and MongoDB) may be running on the host or inside Docker; Docker Compose configuration is provided for convenience but not required.
```
noisersup, line 4: > Docker Compose configuration is provided for convenience but not required.
It's required now to run tests/mongo-tools

w84thesun (commented):
That's a lot of work! Asked about CONTRIBUTING.md change.

CONTRIBUTING.md
noisersup: Yeah, let's mention it, though it's still temporary and should be replaced in another PR

w84thesun (dismissed):

rumyantseva (dismissed):
LGTM.

The current approach works for me locally when I have FerretDB running on my host (port 27017):

- For `diff` tests everything works as before - no `env-up` in dance needed etc 
- For `mongo-tools`: 
    - From the `dance` directory I run `docker compose up mongosh`
    - Then I run `task dance DB=ferretdb TEST=mongo-tools` and it passes!
    - In the logs of my local FerretDB instance I see that it was called and reached.

Taskfile.yaml
```diff
@@ -27,6 +27,11 @@ tasks:
     cmds:
       - docker compose logs --follow
 
+  env-mongosh:
+    desc: "Start mongosh container"
+    cmds:
+      - docker compose up mongosh -d
+
```
noisersup, line 34: Okay, so the solution to all of local instance issues is just a new task (which is now in deps of `task dance`). Thanks @rumyantseva for reminding that it's possible to run a single container from docker-compose :slightly_smiling_face:

rumyantseva: Yeah, the next step would be to not run it at all (as @AlekSi mentioned on Slack).

noisersup: done

AlekSi (approved):
Let's start with that

w84thesun (approved):

PR #256, opened 2022-11-25
Update go driver tests

b1ron:
Closes @253.

### DIFF
--- a/tests/mongo-go-driver.yml
+++ b/tests/mongo-go-driver.yml
@@ -62,7 +62,7 @@ results:
       unexpected_pass: 0
       expected_fail: 21
       expected_skip: 17
-      expected_pass: 667
+      expected_pass: 700
 
     pass:
       - go.mongodb.org/mongo-driver/mongo/TestClient/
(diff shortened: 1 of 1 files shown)


tests/mongo-go-driver.yml
```diff
@@ -132,3 +132,4 @@ results:
       - go.mongodb.org/mongo-driver/examples/documentation_examples/TestDocumentationExamples/RunCommandExamples
       - go.mongodb.org/mongo-driver/examples/documentation_examples/TestDocumentationExamples/StableAPExamples
       - go.mongodb.org/mongo-driver/examples/documentation_examples/TestDocumentationExamples/UpdateExamples
+
```
AlekSi, line 135:
```suggestion
```

AlekSi (commented):
That PR should also update submodule with tests

tests/mongo-go-driver.yml
b1ron: So we are removing the newline now?

AlekSi: The last line of the file should be not empty and end with \n

AlekSi: CleanShot 2022-11-25 at 14 59 46@2x

That means that line does not end with \n

AlekSi (approved):

rumyantseva (approved):
That's great to see some progress with mongo go driver tests!

PR #257, opened 2022-11-28
Tweak Docker Compose configuration

AlekSi:


docker-compose.yml
```diff
@@ -2,60 +2,61 @@
...
     environment:
       # Always UTC+05:45. Set to catch timezone problems.
       - TZ=Asia/Kathmandu
 
+  # for tools; always runs, but no ports are exposed
   mongosh:
     build:
       context: ./build/deps
       dockerfile: mongo6.Dockerfile
+    command: --port 12345 # should not be used
```
noisersup, line 55: I don't understand this line
What's the point of specifing the port and what do you mean by "should not be used"?

docker-compose.yml
```diff
@@ -2,60 +2,61 @@
...
     container_name: dance_mongosh
     restart: unless-stopped
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
rumyantseva, line 59: Hmm, I don't really understand how it should work. I tried it locally, and still don't understand.

So, I have `task env-up` from ferretdb and `task run` from ferretdb. Those are the containers I have at the moment:
```
docker ps --format "table {{.ID}}\t{{.Image}}\t{{.Names}}\t{{.Ports}}\t{{.Networks}}"
CONTAINER ID   IMAGE                NAMES                PORTS                                NETWORKS
7fbf4356a718   ferretdb-mongodb     ferretdb_mongodb     0.0.0.0:37017->27017/tcp             ferretdb_default
f274652069c4   ferretdb-postgres    ferretdb_postgres    0.0.0.0:5432->5432/tcp               ferretdb_default
60118230244a   ferretdb-tigris      ferretdb_tigris      0.0.0.0:8081->8081/tcp               ferretdb_default
67140e712b45   ferretdb-cockroach   ferretdb_cockroach   8080/tcp, 0.0.0.0:26257->26257/tcp   ferretdb_default
```

Now what can I do with dance? I tried `task mongosh DB=ferretdb`, but it gives me the following:
```
task mongosh DB=ferretdb
task: [mongosh] docker-compose exec mongosh mongosh mongodb://ferretdb:27017/ --verbose --eval 'disableTelemetry()' --shell

service "mongosh" is not running container #1
task: Failed to run task "mongosh": exit status 1
```

How it is expected to work?

rumyantseva: Like, should I call `env up` from `dance` and then `task run` from ferretdb?

chilagrow: So this seems to expose localhost to docker. I don't quite know the user case though.

noisersup: @chilagrow It's required if we want to run mongo tools (like `mongodump`) from mongosh container against FerretDB running locally (with `task run`)

docker-compose.yml
noisersup: Oh okay, I guess it's set to remove collidation on 27017 port

docker-compose.yml
chilagrow: Yeah I guessed. But does the step @rumyantseva does align with how you expect to use this @noisersup ?

AlekSi: Please ignore it for now, I will fix it today

PR #260, opened 2022-11-30
Document how to write diff tests

### LINKED ISSUE #214: Document how to write diff tests
As it's not 100% straightforward how to write `diff` tests, it needs to be documented.

Potential items to be documented:
- What diff tests are, when they need to be written
- An example of a diff test (with `MongoDB` and `FerretDB` subtests)
- What changes in yaml files need to be made after new tests are added 
- How to run tests locally and what are expected results (e.g. expected failures for diff tests), how many expected failures and passes should be added
- Document that the main dev cycle in this case is to have ferretdb running on the port 27017 (e.g. through `task run`) and then calling `task dance DB=ferretdb TEST=diff` in dance

In contributing.md of the dance repo.

rumyantseva:
Closes #214. 

Please suggest language, grammar, factual or any other improvements!

Currently dance tests fail because of an issue in FerretDB, the PR to fix it is [here](https://github.com/FerretDB/FerretDB/pull/1602).


CONTRIBUTING.md
```diff
@@ -53,3 +53,80 @@ To build a local image use the `bin/task docker-local` command in the [FerretDB]
...
+```go
+func TestSomeDiff(t *testing.T) {
+    t.Run("SubtestSomething", func(t *testing.T) {
+        t.Run("FerretDB", func(t *testing.T) {})
+        t.Run("MongoDB", func(t *testing.T) {})
+    })
+}
+```
+
+In this example, for FerretDB, the number of passed tests is 1.
```
chilagrow, line 106: I think you meant to add something that fails in the above code?

rumyantseva: I tried to modify an example (introduced it earlier and added the actual logic), please let me know if it looks better now.

CONTRIBUTING.md
```diff
@@ -53,3 +53,92 @@ To build a local image use the `bin/task docker-local` command in the [FerretDB]
...
 As mentioned above, this approach is not recommended.
+
+## How to write diff tests
+
+Diff tests demonstrate how FerretDB differs from MongoDB.
+
+We write diff tests when we can't write "regular" integration or compat tests because FerretDB behaves differently
+from MongoDB in some cases.
+If such differences are visible to users, in addition to diff tests, they must be documented
+in the [FerretDB documentation](https://raw.githubusercontent.com/FerretDB/FerretDB/main/website/docs/diff.md),
```
Fashander, line 64: I think this link will be better "https://docs.ferretdb.io/diff/" here.

Fashander (dismissed):
LGTM

CONTRIBUTING.md
rumyantseva: Yeah, I was thinking which one would be better, and let's add this one.

noisersup (approved):
Nice job with describing a whole process!

chilagrow (dismissed):
:rocket:

CONTRIBUTING.md
```diff
@@ -53,3 +53,92 @@ To build a local image use the `bin/task docker-local` command in the [FerretDB]
...
 
 As mentioned above, this approach is not recommended.
+
+## How to write diff tests
+
+Diff tests demonstrate how FerretDB differs from MongoDB.
+
+We write diff tests when we can't write "regular" integration or compat tests because FerretDB behaves differently
+from MongoDB in some cases.
+If such differences are visible to users, in addition to diff tests, they must be documented
```
AlekSi, line 63: If they are not visible to users, they should not be diff tests

CONTRIBUTING.md
```diff
@@ -53,3 +53,92 @@ To build a local image use the `bin/task docker-local` command in the [FerretDB]
...
+in the [FerretDB documentation](https://docs.ferretdb.io/diff/),
+please take a look at it for the additional context.
+
+Diff tests are located in the `tests/diff` directory and are regular Go tests.
+
+Currently, every diff test or subtest should have at least two subtests - `FerretDB` and `MongoDB`.
```
AlekSi, line 69:
```suggestion
Currently, every diff test or subtest should have two subtests - `FerretDB` and `MongoDB`.
```

CONTRIBUTING.md
```diff
@@ -53,3 +53,92 @@ To build a local image use the `bin/task docker-local` command in the [FerretDB]
...
+```
+
+It works with document validation feature for the `foo$` field key
+and demonstrates how FerretDB validation differs from MongoDB behavior.
+In case of FerretDB a document is considered  invalid if it has a key containing the `$` sign.
+In case of MongoDB such document is considered valid.
+The subtest `Insert` demonstrates this difference: in case of FerretDB we expect to receive a particular error,
+in case of MongoDB we expect to receive no error when we insert such document.
+
+### Configuration
```
AlekSi, line 104: That section is not specific for diff tests and could be generalized

CONTRIBUTING.md
```diff
@@ -53,3 +53,92 @@ To build a local image use the `bin/task docker-local` command in the [FerretDB]
...
+So, for FerretDB the number of passed tests is equal to the number
+of all the `FerretDB` subtests running.
+The number of failed tests is calculated hierarchically: the "parental" test
+and all its subtests that don't match the regular expression `FerretDB$` are considered failed.
+For MongoDB, we have a similar configuration.
+
+In the example above (`TestDollarSign`), for FerretDB, the number of passed tests is 1 (for `FerretDB` subtest).
+The number of failed tests is 3 (`MongoDB` subtest, `Insert`, and finally `TestDollarSign`).
+
+### How it works?
```
AlekSi, line 121: That section is not specific for diff tests and also already present above. Please update section above and remove that one. We should not duplicate information. And we should not only write documentation without revisiting what's already written.

CONTRIBUTING.md
```diff
@@ -53,3 +53,92 @@ To build a local image use the `bin/task docker-local` command in the [FerretDB]
...
+
+The `dance` command reads the configuration file and runs the tests.
+Then it compares
+the actual number of passed and failed tests with the number from the configuration file.
+
+If the numbers are different, dance prints the list of unexpected tests and exits with a non-zero code.
+
+### Developer workflow
+
+A typical developer workflow when working with diff tests is as follows:
```
AlekSi, line 134: The typical developer workflow does not include diff tests at all. And again, we should not duplicate information

CONTRIBUTING.md
rumyantseva: @AlekSi, this is not quite true. We have a diff test for `debugError` command, which we don't mention in the users' diff documentation.

AlekSi: Ok, let's mark it as an exception then. I don't think there will be others

CONTRIBUTING.md
rumyantseva: "A typical developer workflow when working with diff tests" does not include diff tests?

AlekSi: Did we add diff tests during this typical sprint?

rumyantseva: Well, I meant "typical developer workflow when working with diff tests", not "typical developer workflow" in general.  How to rephrase it to highlight that it's only when such tests are included?

w84thesun (approved):
I like it as it is!

CONTRIBUTING.md
rumyantseva: fixed

CONTRIBUTING.md
rumyantseva: fixed

CONTRIBUTING.md
rumyantseva: fixed

CONTRIBUTING.md
rumyantseva: fixed

chilagrow (approved):

noisersup (approved):

CONTRIBUTING.md
```diff
@@ -1,55 +1,164 @@
...
+This document describes how to run and develop dance tools and tests.
+
+## Types of tests
+
+You can find tests in the `tests` directory.
+We support tests of different types:
+
+* `mongo-go-driver` - to test compatibility with the MongoDB Go driver;
+* `diff` - to demonstrate FerretDB-specific differences;
+* `mongo-tools` - to test compatibility with the tools (like dump and restore).
```
AlekSi, line 12:
```suggestion
```

That information should be (and already is) in the README. Please don't ignore it.

It is also not correct. This repository contains two _types_ of tests: unit tests for the dance tool itself, and integration tests that the dance tool runs against FerretDB and MongoDB to track what is needed to be implemented or fixed in FerretDB. The difference between `diff` tests (and now also `mongo-tools` tests) and the rest is who writes them – are they first-party or third-party?

CONTRIBUTING.md
```diff
@@ -1,55 +1,164 @@
...
 
-## Cloning repository
+## How to roll out dance locally
```
AlekSi, line 14: Roll out?

CONTRIBUTING.md
```diff
@@ -1,55 +1,164 @@
...
 
-If you've already cloned it without submodules you can use that command
-to include required submodules.
+## Local environment
+
+The dance tool and tests run on the host; macOS, Linux, and Windows are expected to work.
+Databases under test (FerretDB and MongoDB) may be running on the host or inside Docker.
+Docker Compose configuration is provided for convenience but not required.
+
+In particular, the FerretDB development cycle (fix-compile-run-dance) is faster with it running on the host
```
AlekSi, line 34: The "FerretDB development cycle" does not include running dance tests.

CONTRIBUTING.md
```diff
@@ -1,55 +1,164 @@
...
+as it does not involve Docker image building or PostgreSQL restarts.
+Running FerretDB on the host is recommended for that reason.
+
+## Development cycle
```
AlekSi, line 38:
```suggestion
```

That's the same section, isn't it?

CONTRIBUTING.md
```diff
@@ -1,55 +1,164 @@
...
+
+In principle, `dance` can work with any DB instance running on the port `27017`.
+
+### Running FerretDB
+
+An example of a typical workflow:
+
+* You work with a FerretDB branch on your local machine.
```
AlekSi, line 46: And that's it. Running dance tests is not part of the normal workflow.

AlekSi (commented):
Overall, I feel like this PR changes documentation from what we do to something that we don't actually do or at least should not do.

I will update it.

CONTRIBUTING.md
rumyantseva: It's not a part of the typical workflow of FerretDB, but we are currently in dance. It'a s part of typical workflow when you work with dance.

rumyantseva: And a typical **dev** workflow with dance is that you need to run FerretDB locally, not in a container.

CONTRIBUTING.md
rumyantseva: Ok, then let's move it to README, but let's keep proper formatting - with lists and short sentences.

CONTRIBUTING.md
rumyantseva: Ok, let's call it "clone", but let's keep "How to"

CONTRIBUTING.md
rumyantseva: Then please rephrase it to highlight which dev cycle does include running dance tests. Currently it's not clear, especially for new contributors.

PR #262, opened 2022-12-02
Add tests for `mongodump` and `mongorestore`

### LINKED ISSUE #149: Add tests for `mongorestore` and `mongodump`

noisersup:
Closes #149.

We can't compare BSON files after all.


noisersup:
I've definitely overcomplicated the solution, thanks! :slightly_smiling_face:

chilagrow:
Thank you for the in depth explanation :hugs: 
> Dump file created by FerretDB differs from the mongodb's one. It's because of a different order of documents in collections (NOTE: do we have any plans to fix that?)

Let's create an issue so we can track it. Then we can decide how we prioritise it.

> **To make sure that the dump file doesn't contain invalid data we can:**

I don't quite understand what we can achieve from this. I think this was probably introduced because _something_ wasn't working as expected? If that's the case we should create an issue and test for the specific case instead of _While using FerretDB_, we should have _while we execute specific command which causes invalid data_.

noisersup:
> Would this ticket https://github.com/FerretDB/FerretDB/issues/457 fix it? It's about sorting

It looks like it will! Thanks for noticing that :)

> Is this trying to compare out of order documents in the collection? For example, every single document occures at the same index does this mean we get documents sorted by _id in both database and check sorted collections are the same?

This point was written before change about sorting by `_id`. Now documents in tests are sorted before a dump to make dumps the same.

tests/mongo-tools/dump_restore_test.go
```diff
@@ -21,57 +21,77 @@ import (
...
 
-	// TODO restart existing dump from https://github.com/mcampo2/mongodb-sample-databases instead
-	_, err := db.Collection("test").InsertOne(ctx, bson.D{{"foo", "bar"}})
+	localActualPath := filepath.Join("..", "..", "dumps", "actual")
+	containerActualPath := "/dumps/"
+
+	localExpectedPath := filepath.Join("..", "..", "dumps", "expected")
+	containerExpectedPath := "/sample-dumps/"
+
+	dbName := "sample_geospatial"
```
rumyantseva, line 37: So, in the scope of the current task, we use only this dump. Are there plans to test more dumps in the future, or do we consider this one enough?

tests/mongo-tools/dump_restore_test.go
```diff
@@ -21,57 +21,77 @@ import (
...
 	)
 	require.NoError(t, err)
 
-	// get database state after restore
+	// get database state after restore and compare it
 	actualState := getDatabaseState(t, ctx, db)
 	assert.Equal(t, expectedState, actualState)
+
+	// compare dump files
+	compareDirs(t, filepath.Join(localExpectedPath, dbName), filepath.Join(localActualPath, dbName), "/*.metadata.json")
```
rumyantseva, line 96: Let's maybe add a comment why we ignore metadata.json files.

rumyantseva (commented):
The changes look reasonable to me. 

Asked a couple of things for clarification.

tests/mongo-tools/dump_restore_test.go
noisersup: We needed this dump to test some large real-word dataset. I believe, we don't have any plans for other test cases

rumyantseva: Ok!

tests/mongo-tools/dump_restore_test.go
noisersup: done!

rumyantseva (approved):
LGTM. I see dance tests fail at the moment, but it looks relevant to yesterday's changes, not to the current PR.

chilagrow (approved):

noisersup:
> Merging is blocked
Merging can be performed automatically with 2 approving reviews.

It looks like approving reviews are not counted after merging main to the branch

rumyantseva (approved):
LGTM

chilagrow (approved):

AlekSi (commented):
We also should not put binary files into git repo. Let's add https://github.com/mcampo2/mongodb-sample-databases as a git submodule

tests/mongo-tools/helpers.go
```diff
@@ -82,3 +90,79 @@ func getDatabaseState(t *testing.T, ctx context.Context, db *mongo.Database) map
...
+		difflib.SplitLines(string(content2)),
+	)
+}
+
+// compareDirs compares two directories and their files recursively.
+// It ignores paths based on regex expressions under ignorePathRegs.
+func compareDirs(t *testing.T, dir1, dir2 string, ignorePathRegs ...string) {
+	var regexps []*regexp.Regexp
+	for _, reg := range ignorePathRegs {
+		regexps = append(regexps, regexp.MustCompile(reg))
```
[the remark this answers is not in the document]
noisersup: Yeah, that's right. I don't remember what was my reasoning behind that. Fixed and checked in logs if it matches the path

AlekSi:
> It was added like that because the dump on the submodule is fully in the json format

https://github.com/mcampo2/mongodb-sample-databases/tree/master/dump/sample_geospatial

> I cannot see any advantages of using submodule in this case.

git submodule does not take repository space; it is a separate repository by definition.

tests/mongo-tools/helpers.go
```diff
@@ -82,3 +90,79 @@ func getDatabaseState(t *testing.T, ctx context.Context, db *mongo.Database) map
...
+			if reg.MatchString(path) {
+				return nil
+			}
+		}
+
+		if d.IsDir() {
+			_, err = os.Stat(comparePath)
+			assert.NoError(t, err)
+			return nil
+		}
```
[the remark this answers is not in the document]
noisersup: It's because we don't want to treat directories as files and compare their nonexisting "content". The `WalkDir` function should check for the files of this directory at some iteration.

docker-compose.yml
```diff
@@ -55,7 +55,8 @@ services:
     extra_hosts:
       - "host.docker.internal:host-gateway"
     volumes:
-      - ./dumps:/dumps
+      - ./dumps/actual:/dumps
+      - ./dumps/expected:/sample-dumps
```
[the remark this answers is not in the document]
noisersup: I thought placing all expected and actual dumps into one root directory of the project would be just better than having two additional top level directories

noisersup:
> https://github.com/mcampo2/mongodb-sample-databases/tree/master/dump/sample_geospatial

Thanks!

> git submodule does not take repository space; it is a separate repository by definition.

If you want to run dance tests you still need to fetch whole submodule and as far as I recall there's no easy way to just fetch specific directory from submodule. So we'll need to fetch additional 1GB of data on every CI run

AlekSi:
Correct – fetch, but not store. Fetching data from GitHub on GitHub's CI is fast. Working with a large repository with binary files is slow for both GitHub's CI and us.

noisersup:
I see your point. Let me apply the changes I will re-request review asap to make it merged, thanks!

docker-compose.yml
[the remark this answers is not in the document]
AlekSi: I think having `/dumps/actual` and `/dumps/expected` on both the host and inside the container would simplify things

tests/mongo-tools/helpers.go
```diff
@@ -82,3 +90,79 @@ func getDatabaseState(t *testing.T, ctx context.Context, db *mongo.Database) map
...
+	require.NoError(t, err)
+
+	hash2 := h.Sum(nil)
+
+	// compare hashes of both files
+	if assert.Equal(t, hash1, hash2, "Checksums of following files are different:", file1.Name(), file2.Name()) {
+		return
+	}
+
+	content1, err := io.ReadAll(file1)
```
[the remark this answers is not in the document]
noisersup: done

tests/mongo-tools/helpers.go
```diff
@@ -82,3 +90,79 @@ func getDatabaseState(t *testing.T, ctx context.Context, db *mongo.Database) map
...
+	require.NoError(t, err)
+
+	content2, err := io.ReadAll(file2)
+	require.NoError(t, err)
+
+	difflib.NewMatcher(
+		difflib.SplitLines(string(content1)),
+		difflib.SplitLines(string(content2)),
+	)
```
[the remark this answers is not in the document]
noisersup: done

docker-compose.yml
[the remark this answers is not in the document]
noisersup: done

tests/mongo-tools/dump_restore_test.go
```diff
@@ -21,57 +21,100 @@ import (
...
+	// dump a database
+	mongodump(t, dbName, containerActualPath)
+
+	// get database state after restore and compare it
+	actualState := getDatabaseState(t, ctx, db)
+	assert.Equal(t, expectedState, actualState)
+
+	// compare dump files. Metadata files are not compared because they
+	// contain different uuid field on every dump
+	compareDirs(t, filepath.Join(localExpectedPath, dbName), filepath.Join(localActualPath, dbName), `\\*.metadata.json`)
```
chilagrow, line 90: Do want to check that `metadata.json` exist in both localActualPath and localExpectedPath?

tests/mongo-tools/dump_restore_test.go
```diff
@@ -21,57 +21,100 @@ import (
...
+
+	// restore a database from the expected dump
+	mongorestore(t, dbName, containerExpectedPath)
+
+	// dump a database
+	mongodump(t, dbName, containerActualPath)
+
+	// get database state after restore and compare it
+	actualState := getDatabaseState(t, ctx, db)
+	assert.Equal(t, expectedState, actualState)
```
chilagrow, line 86: We don't need to do this comparison for this task, I think that was our discussion :hugs:

tests/mongo-tools/dump_restore_test.go
noisersup: *.metadata.json is the file containing index definitions and it's not required to perform a dump.
Yes, we could check at least if it exists, because it's actually created on every dump (with a different uuid field), although we don't check how `mongodump` works, but if it works with FerretDB so imo we cand just skip it as we don't even provide the index functionality

tests/mongo-tools/dump_restore_test.go
noisersup: I believe that discussion was focused more on testing if FerretDB omits any data from the external **source dump** generated by MongoDB. We've already merged the database state comparison with the past PRs, so I assumed it's still in a scope.

We aim to compare dumps created by FerretDB, but as binary files are... _just binary files_, the comparison of them is just a matter of saying "Yes they're equal" or "No, they're not". 

Comparing the database state under both of these dumps is just handful for the output purposes.

Let's maybe leave this conversation open to discuss it with everyone if needed :eyes:

w84thesun (approved):
Changes are looking reasonable to me. I thought there might be some sorting problems but it's working, I guess.

tests/mongo-tools/dump_restore_test.go
chilagrow: thanks for the explanation. Yes we ignore index for now. Didn't realise metadata was related to index :see_no_evil:

tests/mongo-tools/dump_restore_test.go
chilagrow: Yeah okie, let's keep it for now.

chilagrow (approved):

tests/mongo-tools/helpers.go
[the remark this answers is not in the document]
AlekSi: Then I guess that should be just
```go
if d.IsDir() {
 return nil
}
```

I don't understand how `WalkDir` could find non-existing directory

AlekSi:
It is not ready for review

tests/mongo-tools/helpers.go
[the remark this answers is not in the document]
noisersup: I think I don't understand. Let's say we have 2 directories that we want to compare: A and B.
```
├── A/
│   ├── boo/
│   └── foo/
│       └── bar.txt
└── B/
    ├── boo/
    └── foo/
        └── bar.txt
```
Walkdir that is running on A/ even when A/boo/ is empty will go through it, so it's worth to check if the B/boo/ also exists.

noisersup:
> It is not ready for review

Why?

AlekSi:
GitHub keeps eating the large comments I wrote with a lousy internet connection; only two went through 😞. See https://github.com/FerretDB/dance/pull/262/files#r1049605371 and https://github.com/FerretDB/dance/pull/262/files#r1049599247. Basically, something weird is going on if the code that could not be correct somehow passes tests. I'm taking a deeper look.

tests/mongo-tools/helpers.go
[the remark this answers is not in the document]
noisersup: First and foremost thanks for a wide explanation :pray:

> It is not clear why we specifically check that file exists, given that os.Open would return the same error.

This change was made only to make the output little bit more readable if file does not exist.

> I'm not sure what we do next. It seems that this code tries to compare files' content – but why, given that we already compared hashes exactly to prevent reading the whole files into memory?

This part of code is called only when checksums are different and it was added only to be more verbose about specific differences in files.

> And then why do we rewind only the first file but not the second?

Nice catch, it doesn't fail because it's not even called, as checksums are the same

tests/mongo-tools/helpers.go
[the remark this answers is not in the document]
chilagrow: Just so I understand, this is about testing _expected_ (_dir1_) having an empty directory but _actual_ (_dir2_) is not existing? I feel this isn't necessary since there is no files inside neither directory, we don't worry about if it exist or not.

What's more interesting is if _expected_ (_dir1_) have an empty directory but _actual_ (_dir2_) has files. But let's not go there in this task for now.

rumyantseva (commented):
Could you please re-request review when the PR is ready?

w84thesun (commented):
Please re-request when ready.

w84thesun (approved):
LGTM

chilagrow (approved):
Great!

AlekSi:
@noisersup PTAL

AlekSi (approved):

tests/mongo-tools/dump_restore_test.go
```diff
@@ -15,63 +15,112 @@
...
+			mongorestore(t, name1, containerTestsRoot, name2)
+			actualCount = getDocumentsCount(t, ctx, db2)
+			assert.Equal(t, tc.documentsCount, actualCount)
 
-	// restore a database based on created dump
-	err = runDockerComposeCommand(
+			compareDatabases(t, ctx, db1, db2)
+
+			// we can't compare bson dump files because `mongodump` queries documents in natural order,
+			// but we don't support it
```
noisersup, line 85: I think I might still don't understand why we are unable to generate same dumps. The "bootstrapping" approach was able to generate 2 same dumps as the order on the same database is preserved. Why we removed it?

tests/mongo-tools/dump_restore_test.go
```diff
@@ -15,63 +15,112 @@
...
 		"mongorestore",
-		"--dir", filepath.Join(containerRoot, db.Name()),
-		"--db", db.Name(),
-		"--verbose",
+		"--verbose=2",
+		"--nsInclude="+db+".*",
+		"--nsFrom="+db+".*",
+		"--nsTo="+newDB+".*",
+		"--objcheck",
+		"--drop",
```
noisersup, line 102: cool! :)

noisersup (commented):
I cannot approve my PR, but LGTM. One question

tests/mongo-tools/dump_restore_test.go
AlekSi: Because it doesn't work for the second test with a slightly bigger database. The fact that the file was exactly the same for a smaller database was a happy accident

PR #265, opened 2022-12-05
Add diff tests for infinity cases

rumyantseva:
See also https://github.com/FerretDB/FerretDB/pull/1622.

### DIFF
--- a/tests/diff.yml
+++ b/tests/diff.yml
@@ -5,8 +5,8 @@ args: ["-timeout=30s", "-shuffle=on", "-race", "./..."]
 results:
   ferretdb:
     stats:
-      expected_fail: 49
-      expected_pass: 20
+      expected_fail: 52
+      expected_pass: 22
 
     pass:
       - regex: FerretDB$
@@ -16,8 +16,8 @@ results:
 
   mongodb:
     stats:
-      expected_fail: 49
-      expected_pass: 20
+      expected_fail: 52
+      expected_pass: 22
 
     pass:
       - regex: MongoDB$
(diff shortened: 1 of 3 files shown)


tests/diff/float_values_test.go
```diff
@@ -29,50 +29,45 @@ func TestFloatValues(t *testing.T) {
 	t.Parallel()
 
 	ctx, db := common.Setup(t)
-
-	t.Run("Insert", func(t *testing.T) {
```
rumyantseva, line 33: This subtest is not needed as we only test inserts anyway.

AlekSi: Well, we should also do validation on updates, findAndModify, etc. Let's keep and add a stub for updates

tests/diff.yml
```diff
@@ -5,22 +5,34 @@ args: ["-timeout=30s", "-shuffle=on", "-race", "./..."]
...
     pass:
       - regex: FerretDB$
 
     fail:
       - not_regex: FerretDB$
 
+    # https://github.com/FerretDB/dance/issues/266
+    skip:
+      - github.com/FerretDB/dance/tests/diff/TestFloatValues/FindAndModify
+      - github.com/FerretDB/dance/tests/diff/TestFloatValues/Update
```
rumyantseva, line 21: @AlekSi this configuration doesn't seem to be correct (as dance fails), but I don't understand why. What did I do wrong?

w84thesun (dismissed):

w84thesun (approved):

chilagrow (approved):

AlekSi (commented):
LGTM

PR #284, opened 2022-12-20
Create the `jstest` runner

### LINKED ISSUE #276: Run MongoDB's jstests with dance
We should make a dance runner for MongoDB's jstests.

Before that, we should find a binary for the legacy `mongo` shell or build it ourselves at https://github.com/FerretDB/deps.

DoD: https://github.com/mongodb/mongo/blob/master/jstests/core/basic1.js runs on CI with dance.

* https://github.com/FerretDB/dance/pull/284

b1ron:
Closes #276.


internal/jstest/jstest.go
```diff
@@ -0,0 +1,103 @@
...
+
+			for _, m := range matches {
+				i := strings.LastIndex(m, "/jstests")
+				m = m[i+1:]
+				files = append(files, m)
+			}
+		}
+	}
+
+	// builds the --eval statement
```
AlekSi, line 51: What we are trying to do there?

b1ron: We are trying to load and execute the files in the shell using `eval` because providing a list of .js files with Docker compose does not seem to work. It will load the file and execute it.

```javascript
---
> load('jstests/core/all.js')
true
> 
```

AlekSi: I think the problem is with shell globs, and it could be solved by Go code. For now, let's focus on one file without globs

internal/jstest/jstest.go
```diff
@@ -0,0 +1,103 @@
...
+		if i == len(files)-1 {
+			break
+		}
+		sb.WriteRune(',')
+	}
+	sb.WriteRune(')')
+	sb.WriteRune('"')
+
+	// try
+	err := runCommand("mongo", uri, sb.String())
```
b1ron, line 66: I just noticed I never provide the actual `--eval` statement...

cmd/dance/main.go
```diff
@@ -119,7 +120,21 @@ func main() {
 			log.Fatal(err)
 		}
 
-		runRes, err := gotest.Run(ctx, dir, config.Args, *vF)
+		var runRes *internal.TestResults
+		switch config.Runner {
+		case "gotest":
+			runRes, err = gotest.Run(ctx, dir, config.Args, *vF)
+		case "jstest":
+			err = jstest.Run(ctx, config.Args)
```
AlekSi, line 128: I think we should make runner return results. We can treat each file/argument as a test

internal/jstest/jstest.go
b1ron: > I think the problem is with shell globs, and it could be solved by Go code. For now, let's focus on one file without globs

Yes that seems to be the problem. I left the logic in there for convenience and passing a single file doesn't break it.

internal/jstest/jstest.go
```diff
@@ -0,0 +1,74 @@
...
+	"strings"
+)
+
+const uri = "mongodb://host.docker.internal:27017"
+
+func Run(ctx context.Context, args []string) error {
+	// TODO https://github.com/FerretDB/dance/issues/20
+	_ = ctx
+
+	// bad
```
AlekSi, line 33: bad? :)

internal/jstest/jstest.go
```diff
@@ -0,0 +1,74 @@
...
+	files := []string{}
+	for _, f := range args {
+		if strings.HasSuffix(f, ".js") {
+			wd, _ := os.Getwd()
+			ff := filepath.Join(wd, "mongo", f)
+			matches, err := filepath.Glob(ff)
```
AlekSi, line 39: Let's remove globs for now

tests/mongo.yml
```diff
@@ -0,0 +1,18 @@
+---
+runner: jstest
+args:
+  # - --quiet
+  - jstests/core/basic1.js
+  # uncaught exception: TypeError: TestData is undefined
+  # - jstests/auth/auth1.js
```
AlekSi, line 7: Please remove commented-out lines

AlekSi (changes requested):

internal/jstest/jstest.go
b1ron: the code :D. fixed.

cmd/dance/main.go
b1ron: I used the `internal.TestResult` type for now just to see if it would work. Should we treat this runner and its output separately and not use the same types? Given that the output is different etc.

b1ron: I think we need to also parse stdout for assertions etc. to have detailed results.

cmd/dance/main.go
```diff
@@ -119,7 +120,17 @@ func main() {
...
 		}
 
-		runRes, err := gotest.Run(ctx, dir, config.Args, *vF)
+		var runRes *internal.TestResults
+		switch config.Runner {
+		case "gotest":
+			runRes, err = gotest.Run(ctx, dir, config.Args, *vF)
+		case "jstest":
+			// TODO return an error too?
+			runRes = jstest.Run(ctx, config.Args)
```
AlekSi, line 129: I think yes, we should return both result and fatal error, just like gotest

internal/jstest/jstest.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+const uri = "mongodb://host.docker.internal:27017"
+
+func Run(ctx context.Context, args []string) *internal.TestResults {
+	// TODO https://github.com/FerretDB/dance/issues/20
+	_ = ctx
+
+	files := []string{}
+	for _, f := range args {
```
AlekSi, line 35: I think we could treat all arguments as js file names, and add `--verbose` and  `--norc` flags to `dockerArgs`, just like in gotest:

https://github.com/FerretDB/dance/blob/ef96d6eadf48523c482aec0f3cf5331840382ea9/internal/gotest/gotest.go#L32

internal/jstest/jstest.go
```diff
@@ -0,0 +1,77 @@
...
+
+	ts := &internal.TestResults{}
+	ts.TestResults = make(map[string]internal.TestResult)
+	for _, f := range files {
+		err := runCommand("mongo", uri, f)
+		ts.TestResults[f] = internal.TestResult{}
+		if err != nil {
+			ts.TestResults[f] = internal.TestResult{
+				Status: internal.Fail,
+				Output: "whatever..",
```
AlekSi, line 49: I think we should capture stdout and stderr for both `err == nil` and `err != nil` with [`os/exec.Cmd.CombinedOutput`](https://pkg.go.dev/os/exec#Cmd.CombinedOutput) and return it

cmd/dance/main.go
AlekSi: No, runners should share result types and config type, that's the point.

I think there is no need to parse stdout/stderr yet, just return it

w84thesun (commented):
Looks like this one is still in progress.

rumyantseva (commented):
Please re-request review when it's ready (also feel free to ping anyone from the team if there is something unclear)

b1ron:
There seems to be a lot of jobs that are failing now.

AlekSi (approved):

PR #293, opened 2022-12-27
Add globs in `jstest` runner

### LINKED ISSUE #289: Add support for globs in `jstest` runner
```yaml
runner: jstest
args:
  - tests/mongo/jstests/core/basic1.js
  - tests/mongo/jstests/core/basic2.js
```

It should be possible to replace that configuration with:
```yaml
runner: jstest
args:
  - tests/mongo/jstests/core/basic*.js
```

There is no shell, so we should do that in Go, using https://pkg.go.dev/path/filepath#Glob and similar functions.

b1ron:
Closes #289.

### DIFF
--- a/internal/jstest/jstest.go
+++ b/internal/jstest/jstest.go
@@ -34,16 +34,12 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 	files := []string{}
 
 	for _, f := range args {
-		i := strings.LastIndex(f, dir)
-
-		matches, err := filepath.Glob(f[i:])
+		matches, err := filepath.Glob(f)
 		if err != nil {
 			return nil, err
 		}
 
-		for _, m := range matches {
-			files = append(files, filepath.Join("tests", m))
-		}
+		files = append(files, matches...)
 	}
 
 	for _, testName := range files {
--- a/internal/jstest/jstest.go
+++ b/internal/jstest/jstest.go
@@ -17,7 +17,9 @@ package jstest
 import (
 	"context"
 	"log"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/FerretDB/dance/internal"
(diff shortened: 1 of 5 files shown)


internal/jstest/jstest.go
```diff
@@ -29,7 +31,28 @@ func Run(ctx context.Context, args []string) (*internal.TestResults, error) {
...
+
+	files := []string{}
+
+	for _, f := range args {
+		dir, err := os.Getwd()
+		if err != nil {
+			return nil, err
+		}
+
+		matches, err := filepath.Glob(filepath.Join(dir, f))
```
AlekSi, line 43:
```suggestion
		matches, err := filepath.Glob(dir, f)
```

The current directory is implied

internal/jstest/jstest.go
```diff
@@ -29,7 +31,28 @@ func Run(ctx context.Context, args []string) (*internal.TestResults, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		for _, m := range matches {
+			i := strings.LastIndex(m, "/tests/")
```
AlekSi, line 49: Why we need that?

tests/mongo.yml
```diff
@@ -1,15 +1,14 @@
...
 
 results:
   ferretdb:
     stats:
       expected_fail: 2
-      expected_pass: 1
+      expected_pass: 7
     fail:
       # https://github.com/FerretDB/FerretDB/issues/161
       - tests/mongo/jstests/core/basic1.js
```
AlekSi, line 14:
```suggestion
      - mongo/jstests/core/basic*.js
```

I think it makes sense to use the same path format as in `args`

internal/jstest/jstest.go
b1ron: Fixed

tests/mongo.yml
b1ron: Fixed

internal/jstest/jstest.go
b1ron: Fixed.

b1ron:
I mounted the volume under `/mongo` to make the code cleaner. Might that cause problems with the build given that we clone the repo?

internal/jstest/jstest.go
```diff
@@ -18,18 +18,31 @@ import (
...
 	"log"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/FerretDB/dance/internal"
 )
 
-func Run(ctx context.Context, args []string) (*internal.TestResults, error) {
+func Run(ctx context.Context, dir string, args []string) (*internal.TestResults, error) {
```
AlekSi, line 27: `dir` is not used

AlekSi:
I think the problem is that, unlike in `gotest` runner, `cmd.Dir` is never set

internal/jstest/jstest.go
b1ron: Yes https://github.com/FerretDB/dance/pull/293/commits/a5ac0416f29843f7a6e037c61533c696b7772d69 added confusion, my bad. I removed `dir` and it works.

b1ron:
> I think the problem is that, unlike in gotest runner, cmd.Dir is never set

As discussed this is not needed as we are in the calling process's current directory and everything works from this directory.

AlekSi:
Added `cmd.Dir = dir` just for consistency

AlekSi (approved):

PR #296, opened 2022-12-28
Add tests for `mongoimport` and `mongoexport`

### LINKED ISSUE #148: Add tests for `mongoimport` and `mongoexport`
We should run `mongoimport` and `mongoexport` to ensure we stay compatible with FerretDB. Let's add a new set of tests that would run binaries, not use their tests.

* https://www.mongodb.com/docs/database-tools/mongoimport/
* https://www.mongodb.com/docs/database-tools/mongoexport/

## DoD:

- [x] Port the existing dump-restore test to import/export
- [x] Flags to check: `sort`, maybe some others if it makes sense

### What tests need to be added and where?
As for mongodump task (#149), we should validate if FerretDB works with mongoimport and mongoexport tools.
It should be added in existing `tests/mongo-tools` directory.

### What tests should check/compare?
- import the json file from mongodb-sample-databases
- export it to another json file
- import it again
- compare the database states after first and second import

### Not in scope
- comparing the files
- any things related to mongo tools testing (as testing various filetypes, etc.)
- any things related to functionality testing (as omiting data and everything that we currently test with our integration tests for specific commands that are used by mongo tooling)

### Additional informations

From my investigation on local machine, it seems that FerretDB's json file contains the same fields as the dump in the sample databases repo and it only differs by their order. 
After using mongoexport's `--sort` flag, `diff` util shows that the dumps are exactly the same, so probably no bootstrapping is needed. 

It looks like indexes are not the problem in this case, but I'm still not 100% sure.

noisersup:
Closes #148.

### DIFF
--- a/tests/mongo-tools/dump_restore_test.go
+++ b/tests/mongo-tools/dump_restore_test.go
@@ -26,8 +26,8 @@ func TestDumpRestore(t *testing.T) {
 	t.Parallel()
 
 	containerSourceRoot := "/dumps/mongodb-sample-databases/dump"
-	localTestsRoot := filepath.Join("..", "..", "dumps", "mongoexport_tests")
-	containerTestsRoot := "/dumps/mongoexport_tests"
+	localTestsRoot := filepath.Join("..", "..", "dumps", "mongodump_tests")
+	containerTestsRoot := "/dumps/mongodump_tests"
 
 	//nolint:vet // for readability
 	type testCase struct {
--- a/tests/mongo-tools/dump_restore_test.go
+++ b/tests/mongo-tools/dump_restore_test.go
@@ -26,8 +26,8 @@ func TestDumpRestore(t *testing.T) {
 	t.Parallel()
 
 	containerSourceRoot := "/dumps/mongodb-sample-databases/dump"
-	localTestsRoot := filepath.Join("..", "..", "dumps", "tests")
-	containerTestsRoot := "/dumps/tests"
+	localTestsRoot := filepath.Join("..", "..", "dumps", "mongoexport_tests")
+	containerTestsRoot := "/dumps/mongoexport_tests"
 
 	//nolint:vet // for readability
 	type testCase struct {
--- a/tests/mongo-tools/export_import_test.go
+++ b/tests/mongo-tools/export_import_test.go
@@ -29,9 +29,9 @@ func TestExportImport(t *testing.T) {
 	localTestsRoot := filepath.Join("..", "..", "dumps", "tests")
 
 	type testCase struct {
-		coll           string
-		db             string
-		documentsCount int // document count
+		coll           string // target collection
+		db             string // target database
+		documentsCount int    // document count
 	}
 
 	for name, tc := range map[string]testCase{
@@ -40,7 +40,7 @@ func TestExportImport(t *testing.T) {
 			db:             "sample_geospatial",
 			documentsCount: 11095,
 		},
-		"accounts": {
+		"Accounts": {
 			coll:           "accounts",
 			db:             "sample_analytics",
 			documentsCount: 1746,
@@ -49,8 +49,8 @@ func TestExportImport(t *testing.T) {
 
 		name, tc := name, tc
 		t.Run(name, func(t *testing.T) {
-			dbName1 := name + "_dump1"
-			dbName2 := name + "_dump2"
+			dbName1 := databaseName(t) + "_dump1"
+			dbName2 := databaseName(t) + "_dump2"
 
 			sourceFile := filepath.Join(localSourceRoot, tc.db, tc.coll+".json")
 			testFile := filepath.Join(localTestsRoot, dbName1, tc.coll+".json")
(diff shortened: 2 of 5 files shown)


tests/mongo-tools/export_import_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExportImport(t *testing.T) {
+	t.Parallel()
+
+	containerSourceRoot := "/dumps/mongodb-sample-databases/"
+	containerTestsRoot := "/dumps/tests"
+	localTestsRoot := filepath.Join("..", "..", "dumps", "tests")
```
noisersup, line 30: Thanks to the usage of the testname in the database name it's easy to store both mongorestore and mongodump tests inside the same directory

tests/mongo-tools/helpers.go
```diff
@@ -49,6 +49,22 @@ func setup(tb testing.TB) (context.Context, *mongo.Client) {
 	return context.Background(), client
 }
 
+// databaseName returns a stable database name for that test.
+func databaseName(tb testing.TB) string {
+	tb.Helper()
```
noisersup, line 54: I've seen that we removed the common helpers with the last PR, so I'm assuming that code repetition is ok in that case. If not, let's extract it to some top level helpers file

tests/mongo-tools/export_import_test.go
noisersup: It seems it's not, will fix that. But all other things can be reviewed

rumyantseva (approved):
LGTM, the logic mentioned in the issue is implemented.

tests/mongo-tools/dump_restore_test.go
```diff
@@ -26,8 +26,8 @@ func TestDumpRestore(t *testing.T) {
 	t.Parallel()
 
 	containerSourceRoot := "/dumps/mongodb-sample-databases/dump"
-	localTestsRoot := filepath.Join("..", "..", "dumps", "tests")
-	containerTestsRoot := "/dumps/tests"
+	localTestsRoot := filepath.Join("..", "..", "dumps", "mongoexport_tests")
+	containerTestsRoot := "/dumps/mongoexport_tests"
```
AlekSi, line 30: I think `mongoexport_tests` and `mongodump_tests` are swapped between there and export_import_test.go

tests/mongo-tools/export_import_test.go
```diff
@@ -0,0 +1,119 @@
...
+		"Accounts": {
+			coll:           "accounts",
+			db:             "sample_analytics",
+			documentsCount: 1746,
+		},
+	} {
+		name, tc := name, tc
+		t.Run(name, func(t *testing.T) {
+			dbName1 := databaseName(t) + "_dump1"
+			dbName2 := databaseName(t) + "_dump2"
```
AlekSi, line 53:
```suggestion
			dbName1 := tc.db + "_export1"
			dbName2 := tc.db + "_export2"
```

When I worked on previous tests, I found testname-based database names more confusing

tests/mongo-tools/helpers.go
AlekSi: I think we don't need that helper; see the other comment

tests/mongo-tools/dump_restore_test.go
noisersup: whoops... nice catch!

tests/mongo-tools/export_import_test.go
noisersup: I would suggest to use database and collection names here. This will create the seperate database for every test case, so it will be possible to run parallel tests on 2 collections from the same "dump"
```go
dbName1 := fmt.Sprintf("%s_%s_export1", tc.db, tc.coll)
dbName2 := fmt.Sprintf("%s_%s_export2", tc.db, tc.coll)
```

noisersup:
Diff tests started failing, from a quick look I assume that we changed the error type on the [ordered inserts task](https://github.com/FerretDB/FerretDB/pull/1719)?

rumyantseva:
@noisersup this is the reason for tests to start failing - dance doesn't check errors correctly -https://github.com/FerretDB/dance/pull/301. 

See also: https://ferretdb.slack.com/archives/C02NWN1K3MJ/p1672333409614989

rumyantseva:
I think the simplest way is to temporarily comment the failing tests, so you can proceed with this PR.

noisersup:
> I think the simplest way is to temporarily comment the failing tests, so you can proceed with this PR.

There's quiet a lot of insert tests that need to be commented to do that. We could just use `t.Skip` on them, though it would require to change our diff.yml regex because currently it match these insert tests as expected to pass.

Why can't we just ignore the diff here, as it's the error on the main branch and not on the PR branch?

rumyantseva:
Yeah, sounds reasonable!

rumyantseva (approved):
LGTM

rumyantseva (approved):

AlekSi (approved):

PR #303, opened 2022-12-30
Fix type of error that need to be checked for `insert`

rumyantseva:
In the case of the `insert` operation, the error that should be returned is `mongo.WriteException`, not `mongo.CommandError`.

It was fixed in FerretDB here - https://github.com/FerretDB/FerretDB/pull/1673, and diff tests need to be fixed too.


b1ron (approved):
Looks great.

AlekSi (approved):

PR #310, opened 2023-01-10
`$mul` returns validation error on unsupported +INF result

chilagrow:
Closes FerretDB/FerretDB#625. 

This tests differences introduced in FerretDB/FerretDB#1760.

* `+INF` upon $mul operator returns error
* `-0` negative zero is converted to `0` positive zero everywhere including insert and update operation.

### DIFF
--- a/tests/diff/float_values_test.go
+++ b/tests/diff/float_values_test.go
@@ -100,9 +100,10 @@ func TestFloatValues(t *testing.T) {
 				insert: bson.D{{"_id", "number"}, {"v", int32(42)}},
 				update: bson.D{{"$mul", bson.D{{"v", math.MaxFloat64}}}},
 				expected: mongo.CommandError{
-					Code:    2,
-					Name:    "BadValue",
-					Message: `invalid value: { "v": +Inf } (infinity values are not allowed)`,
+					Code: 2,
+					Name: "BadValue",
+					Message: `update produces invalid value: { "v": +Inf }` +
+						` (update operations that produce infinity values are not allowed)`,
 				},
 			},
 		} {
--- a/tests/diff/float_values_test.go
+++ b/tests/diff/float_values_test.go
@@ -93,7 +93,7 @@ func TestFloatValues(t *testing.T) {
 		expected := math.Copysign(0.0, +1)
 
 		// testify require equates -0 == 0 so this passes for both -0 and 0.
-		require.Equal(t, 0, actual)
+		require.Equal(t, expected, actual)
 
 		t.Run("FerretDB", func(t *testing.T) {
 			t.Parallel()
@@ -104,7 +104,7 @@ func TestFloatValues(t *testing.T) {
 		t.Run("MongoDB", func(t *testing.T) {
 			t.Parallel()
 
-			require.Equal(t, math.Signbit(expected), math.Signbit(actual))
+			require.NotEqual(t, math.Signbit(expected), math.Signbit(actual))
 		})
 	})
 
@@ -225,7 +225,7 @@ func TestFloatValues(t *testing.T) {
 				t.Run("MongoDB", func(t *testing.T) {
 					t.Parallel()
 
-					require.Equal(t, math.Signbit(expected), math.Signbit(actual))
+					require.NotEqual(t, math.Signbit(expected), math.Signbit(actual))
 				})
 			})
 		}
--- a/tests/diff/float_values_test.go
+++ b/tests/diff/float_values_test.go
@@ -44,15 +44,6 @@ func TestFloatValues(t *testing.T) {
 						`$db: "testfloatvalues", documents: [ { _id: "1", foo: nan.0 } ] } with: NaN is not supported`,
 				},
 			},
-			"NegativeZero": {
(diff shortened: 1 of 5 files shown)


tests/diff/float_values_test.go
```diff
@@ -85,4 +85,50 @@ func TestFloatValues(t *testing.T) {
...
+			expected mongo.CommandError
+		}{
+			"MulMaxFloat64": {
+				filter: bson.D{{"_id", "number"}},
+				insert: bson.D{{"_id", "number"}, {"v", int32(42)}},
+				update: bson.D{{"$mul", bson.D{{"v", math.MaxFloat64}}}},
+				expected: mongo.CommandError{
+					Code:    2,
+					Name:    "BadValue",
+					Message: `invalid value: { "v": +Inf } (infinity values are not allowed)`,
```
rumyantseva, line 105: Ok, after looking at this case I'm not sure we should behave differently :)
"Infinity values are not allowed" is for the case when an input document has infinity in one of the fields. This case is about the result of the operation. 
So, we either should support it and return infinity (if it's possible in our code base), or it should be a different diff.md line for this documentation - https://docs.ferretdb.io/diff/. And I expect the error to be more clear that the result of the operation gives infinity.

rumyantseva: Example: we consider a filter with infinity acceptable, e.g. something like `{v: {$lt: +Inf}}` should be a supported input. So I wonder what should be our behaviour if Inifinity is in the output 🤔

w84thesun: We probably don't want to store infinity values, only use them for comparison. So if we get an infinity value as the operation result we might want to return an error describing that behavior.

w84thesun (commented):
Didn't see from the main PR that infinity error output. We probably want to change the error description.

chilagrow:
> Didn't see from the main PR that infinity error output. We probably want to change the error description.

@w84thesun the logic was already in place we use the same validator as insert doc and I didn't change it https://github.com/FerretDB/FerretDB/blob/main/internal/types/document_validation.go#L142.

I updated error message a bit more specific to update.

tests/diff/float_values_test.go
```diff
@@ -85,4 +85,51 @@ func TestFloatValues(t *testing.T) {
...
+		}{
+			"MulMaxFloat64": {
+				filter: bson.D{{"_id", "number"}},
+				insert: bson.D{{"_id", "number"}, {"v", int32(42)}},
+				update: bson.D{{"$mul", bson.D{{"v", math.MaxFloat64}}}},
+				expected: mongo.CommandError{
+					Code: 2,
+					Name: "BadValue",
+					Message: `update produces invalid value: { "v": +Inf }` +
+						` (update operations that produce infinity values are not allowed)`,
```
rumyantseva, line 106: Yeah, this message looks clear to me! 👍

rumyantseva (dismissed):
LGTM

noisersup (dismissed):
Cool!

w84thesun (approved):
👍

b1ron (approved):

tests/diff/float_values_test.go
```diff
@@ -76,6 +67,47 @@ func TestFloatValues(t *testing.T) {
...
+		t.Run("FerretDB", func(t *testing.T) {
+			t.Parallel()
+
+			require.Equal(t, math.Signbit(expected), math.Signbit(actual))
+		})
+
+		t.Run("MongoDB", func(t *testing.T) {
+			t.Parallel()
+
+			require.NotEqual(t, math.Signbit(expected), math.Signbit(actual))
```
rumyantseva, line 107: I believe you chose `require.NotEqual` over `require.Equal` on purpose, but I don't really understand why... Doesn't `require.Equal` look more readable? 

(Not a request for change, let's see what others think)

rumyantseva (commented):
The necessary tests are added, I left a comment about our previous agreement between diff.md and diff tests.

Also, as we changed the logic for `-0` not only for insert but also for update, we need to add one test for update with `-0`.

tests/diff/float_values_test.go
```diff
@@ -76,6 +67,47 @@ func TestFloatValues(t *testing.T) {
 		}
 	})
 
+	t.Run("InsertNegativeZero", func(t *testing.T) {
```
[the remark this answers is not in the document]
chilagrow: Thanks for pointing this out :pray:

tests/diff/float_values_test.go
chilagrow: You are right, I think it's easier to read and actually assert the equal value instead of checking notEqual. I've changed to below

```
		t.Run("FerretDB", func(t *testing.T) {
			t.Parallel()

			require.Equal(t, math.Signbit(math.Copysign(0.0, +1)), math.Signbit(actual))
		})

		t.Run("MongoDB", func(t *testing.T) {
			t.Parallel()

			require.Equal(t, math.Signbit(math.Copysign(0.0, -1)), math.Signbit(actual))
		})
```

rumyantseva (approved):
LGTM

w84thesun (approved):
👍

noisersup (approved):

Issue #304, opened 2023-01-03
Fix permission issue for parallel shell execution


b1ron:
[updatef.js](https://github.com/mongodb/mongo/blob/master/jstests/core/updatef.js) invokes the shell from inside itself but encounters a permission issue in the container:

```
        {"t":{"$date":"2023-01-03T21:22:16.690Z"},"s":"I",  "c":"-",        "id":22810,   "ctx":"js","msg":"shell: Started program","attr":{"pid":"12","port":-1,"argv":["/mongo","--host","host.docker.internal:27017","--eval","db = db.getSiblingDB('test');for( i=0; i < 100; ++i ) { db.jstests_updatef.renameCollection( 'jstests_updatef_' ); db.jstests_updatef_.renameCollection( 'jstests_updatef' ); }"]}}
        sh12| Unable to start program /mongo: Permission denied
        Error: [0] != [255] are not equal : encountered an error in the parallel shell :
        doassert@src/mongo/shell/assert.js:20:14
        assert.eq@src/mongo/shell/assert.js:180:17
        startParallelShell/<@src/mongo/shell/servers_misc.js:181:20
        @mongo/jstests/core/updatef.js:31:1
        failed to load: mongo/jstests/core/updatef.js
        exiting with code -3
```

AlekSi:
@b1ron that should be fixed by https://github.com/FerretDB/deps/commit/592a1e4b3822216fc2f237f9c5497c0018dc9451 that is already merged. Please check

b1ron:
Fixed by https://github.com/FerretDB/deps/commit/592a1e4b3822216fc2f237f9c5497c0018dc9451.

PR #317, opened 2023-01-12
Fix `jstest` for latest legacy-mongo-shell

b1ron:
Due to https://github.com/FerretDB/deps/commit/592a1e4b3822216fc2f237f9c5497c0018dc9451 we cannot mount the `mongo` submodule at `/mongo`. This PR fixes that issue and prevents tests from breaking.


internal/jstest/jstest.go
```diff
@@ -61,6 +64,9 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 			}
 		}
 
+		// to avoid unexpected results
+		testName = strings.TrimPrefix(testName, prefix+"/")
```
AlekSi, line 68: In general, that should be done using `path/filepath` package, not `path` or `strings`.

But I wonder if we could update mongo.yml instead?

b1ron: The only other solution I could think of is to chdir.

PTAL.

tests/mongo.yml
```diff
@@ -1,97 +1,92 @@
...
-      - mongo/jstests/core/updateb.js
-      - mongo/jstests/core/updatel.js
+      - tests/mongo/jstests/core/updateb.js
+      - tests/mongo/jstests/core/updatel.js
 
       # https://github.com/FerretDB/FerretDB/issues/1744
-      - mongo/jstests/core/updatee.js
-
-      # https://github.com/FerretDB/dance/issues/304
-      - mongo/jstests/core/updatef.js
```
chilagrow, line 85: Has this been removed by accident? :thinking:

tests/mongo.yml
```diff
@@ -1,97 +1,92 @@
...
       # 5. Document restrictions:
       # document keys must not contain $ or . signs;
-      - mongo/jstests/core/updateh.js
+      - tests/mongo/jstests/core/updateh.js
   mongodb:
     stats:
-      expected_fail: 1
-      expected_pass: 53
+      expected_fail: 0
+      expected_pass: 54
```
chilagrow, line 91: These numbers shouldn't change right?

chilagrow (commented):
Just a question about removed tests.

tests/mongo.yml
AlekSi: They should, see removed lines below

internal/jstest/jstest.go
AlekSi: No, I'm not sure what is the best solution, but `chdir` is no-go

tests/mongo.yml
b1ron: updatef.js was removed because https://github.com/FerretDB/dance/issues/304 was fixed.

internal/jstest/jstest.go
```diff
@@ -34,6 +35,12 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 
 	filesM := make(map[string]struct{})
 
+	cwd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+
+	os.Chdir("../")
```
rumyantseva, line 43: I think, in principle, `os.Chdir` can return an error. Do we ignore it on purpose?

Also, I see we use `../` as path, I expect it might not work on Windows, are we ok with it? 

Also, I'm not sure we this command is called here. Why do we need to call it?

rumyantseva (commented):
Asked a question

internal/jstest/jstest.go
b1ron: We change to the parent directory so we can match files using globs and have the filenames be correct when passed to the container. The filenames also need to match so we don't fail when we check the diff. 

Maybe I should put this back into draft so I can find a simpler solution? cc. @AlekSi.

b1ron:
I found an easy solution. Will push it soon.

rumyantseva (approved):
LGTM

chilagrow (approved):

w84thesun (approved):

AlekSi (approved):

PR #351, opened 2023-02-27
Add `env-repl` to Taskfile

b1ron:
Moved to dance.

This PR adds a task to create a single-node replica set. This is useful for testing the oplog and change streams. I opted to create separate containers but we could just convert all instances to a replica set if that's better. However, that would add more log output by introducing [REPL](https://www.mongodb.com/docs/manual/reference/log-messages/#mongodb-data-REPL) messages.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -1,6 +1,9 @@
 ---
 version: "3"
 
+vars:
+  REPLSET: postgres mongodb_replset init_mongodb_replset
+
 tasks:
   init-tools:
     dir: tools
@@ -40,9 +43,17 @@ tasks:
     cmds:
       - docker compose build --pull
 
+  env-repl:
+    desc: "A minimal environment for testing a single-node replica set"
+    cmds:
+      - >
+        docker compose up --remove-orphans --renew-anon-volumes --timeout=0 --detach --build {{.REPLSET}}
+      - task: env-logs
+
   env-logs:
     cmds:
-      - docker compose logs --no-color
+      - docker compose ps --all
+      - docker compose logs --follow
 
   env-down:
     desc: "Stop environment"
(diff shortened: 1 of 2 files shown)


Taskfile.yaml
```diff
@@ -40,9 +43,17 @@ tasks:
     cmds:
       - docker compose build --pull
 
+  env-repl:
+    desc: "A minimal environment for testing a single-node replica set"
```
AlekSi, line 47: Let's add verbs

```suggestion
  env-up-repl:
    desc: "Start a minimal environment for testing a single-node replica set"
```

AlekSi (approved):

AlekSi (approved):

docker-compose.yml
```diff
@@ -75,6 +75,35 @@ services:
...
+  # we need a separate container to run rs.initiate()
+  init_mongodb_replset:
+    build:
+      context: ./build/deps
+      dockerfile: ${MONGO_DOCKERFILE:-mongo6}.Dockerfile
+    container_name: init_mongodb_replset
+    depends_on:
+      - mongodb_replset
+    restart: "no"
+    entrypoint: ["bash", "-c", "sleep 10 && mongosh --host mongodb_replset:47017 --eval 'rs.initiate();'"]
```
rumyantseva, line 105: I see you do `sleep 10` to ensure that `mongodb_replset` is ready. In principle, a better approach would be to configure [startup order](https://docs.docker.com/compose/startup-order/) with some sort of readiness check. However, I don't have a quick way to add such a readiness check to this config, so if `sleep 10` is enough, we can keep it for now.

rumyantseva (approved):

noisersup (approved):

chilagrow (approved):

w84thesun (approved):

b1ron:
We can't merge until https://github.com/FerretDB/dance/pull/350.

Issue #295, opened 2022-12-28
Ensure that we can load files from within `mongo` shell


b1ron:
When we run `jstests/core/array_comparison_correctness.js` it fails because it cannot load `jstests/aggregation/extras/utils.js`:

```
Unexpectedly failed tests:
mongo/jstests/core/array_comparison_correctness.js:
        MongoDB server version: 6.0.3
        {"t":{"$date":"2022-12-28T12:44:37.797Z"},"s":"E",  "c":"-",        "id":22779,   "ctx":"js","msg":"file [{filename}] doesn't exist","attr":{"filename":"jstests/aggregation/extras/utils.js"}}
```

We should ensure we can load all files. I believe we need to keep [`WORKDIR`](https://github.com/FerretDB/deps/blob/main/legacy-mongo-shell/Dockerfile#L29) as `/mongo`.

PR #366, opened 2023-03-16
Add more jstests

### LINKED ISSUE #363: Enable more jstests
We should enable as much as possible. #322 may need to be done first depending on how many we add.

b1ron:
Closes #363.

We need #360 merged before we can merge this.

I will be adding a lot more and creating issues as needed.


b1ron:
#295 will need to be done and now is probably a good time to do so. I noticed two cursor related tests that rely on loading lib files and there's most likely more to encounter as we enable more tests.

AlekSi (approved):

PR #368, opened 2023-03-17
Tweak code owners and dependabot config

AlekSi:


b1ron (approved):
Looks good!

.github/CODEOWNERS
```diff
@@ -6,9 +6,10 @@
 /cmd/                @FerretDB/leads @FerretDB/specops
 /internal/           @FerretDB/leads @FerretDB/specops
 
+/tests/              @b1ron
```
rumyantseva, line 9: Is it enough to only have @b1ron?

rumyantseva (approved):

.github/CODEOWNERS
AlekSi: It is less confusing this way for us

rumyantseva: 👌

PR #376, opened 2023-03-20
Optimize `jstest` runner

### LINKED ISSUE #322: Optimize the `jstest` runner
Currently we spin up a container for each file we run, which doesn't scale. We could, in theory build an `eval` statement and load each file which would only require a single container. This assumes tests are self contained and do not depend on one another.

b1ron:
Closes #322.

### DIFF
--- a/internal/jstest/jstest.go
+++ b/internal/jstest/jstest.go
@@ -54,9 +54,9 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 	}
 
 	type item struct {
-		f   string
-		out []byte
-		err error
+		file string
+		out  []byte
+		err  error
 	}
 
 	var volume = "tests"
@@ -65,7 +65,7 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 	for _, f := range files {
 		go func(f string) {
 			var it item
-			it.f = f
+			it.file = f
 			it.out, it.err = runCommand(dir, "mongo", filepath.Join(volume, f))
 			ch <- it
 		}(f)
@@ -80,14 +80,14 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 		}
 
 		if it.err != nil {
-			res.TestResults[it.f] = internal.TestResult{
+			res.TestResults[it.file] = internal.TestResult{
 				Status: internal.Fail,
 				Output: string(it.out),
 			}
 			continue
 		}
 
-		res.TestResults[it.f] = internal.TestResult{
+		res.TestResults[it.file] = internal.TestResult{
 			Status: internal.Pass,
 			Output: string(it.out),
 		}
--- a/internal/jstest/jstest.go
+++ b/internal/jstest/jstest.go
@@ -53,28 +53,38 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 		TestResults: make(map[string]internal.TestResult),
 	}
 
-	volume := "tests"
+	var volume = "tests"
+
+	ch := make(chan struct{}, len(files))
 	for _, testName := range files {
-		output, err := runCommand(dir, "mongo", filepath.Join(volume, testName))
-		if err != nil {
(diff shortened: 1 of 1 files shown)


b1ron:
The linter is complaining about the type assertion on the error.

internal/jstest/jstest.go
```diff
@@ -53,27 +53,44 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
-				return nil, err
+	type item struct {
+		file string
+		err  error
+		out  []byte
+	}
+
+	var volume = "tests"
+
+	ch := make(chan item, len(files))
```
AlekSi, line 64:
```suggestion
	ch := make(chan *item, len(files))
```

To pass only a pointer, not the whole value

internal/jstest/jstest.go
```diff
@@ -53,27 +53,44 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
+
+	for _, f := range files {
+		go func(f string) {
+			var it item
+			it.file = f
```
AlekSi, line 69:
```suggestion
			it := &item{
			  file: f,
			}
```

internal/jstest/jstest.go
```diff
@@ -53,27 +53,44 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
+			it.out, it.err = runCommand(dir, "mongo", filepath.Join(volume, f))
+			ch <- it
+		}(f)
+	}
+
+	for range files {
```
AlekSi, line 75:
```suggestion
	for it := range ch {
```

AlekSi (commented):
Please use `sync.WaitGroup` to wait for goroutines. And also please run `task lint`

b1ron:
@AlekSi linter is working locally but still getting this [failure](https://github.com/FerretDB/dance/actions/runs/4477690547/jobs/7869535097#step:18:98) on CI. PTAL.

AlekSi:
Please run `task fmt`

internal/jstest/jstest.go
```diff
@@ -53,27 +55,56 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
 
-		if err == nil {
-			res.TestResults[testName] = internal.TestResult{
-				Status: internal.Pass,
-				Output: string(output),
+	// closer
+	go func() {
+		wg.Wait()
+		close(ch)
+	}()
```
AlekSi, line 87:
```suggestion
	wg.Wait()

	close(ch)
```

No need to use a goroutine there

internal/jstest/jstest.go
```diff
@@ -53,27 +55,56 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
+
+	for it := range ch {
+		if it.err != nil {
+			var exitErr *exec.ExitError
+			if !errors.As(it.err, &exitErr) {
+				return nil, it.err
 			}
+		}
 
+		if it.err != nil {
```
AlekSi, line 97: It is the same check as above

internal/jstest/jstest.go
b1ron: Nice!

internal/jstest/jstest.go
b1ron: Nice catch!

internal/jstest/jstest.go
```diff
@@ -53,27 +55,52 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
 	volume := "tests"
-	for _, testName := range files {
-		output, err := runCommand(dir, "mongo", filepath.Join(volume, testName))
-		if err != nil {
-			if _, ok := err.(*exec.ExitError); !ok {
-				return nil, err
+
+	ch := make(chan *item, len(files))
+
+	var wg sync.WaitGroup
```
AlekSi, line 68: We run all files in parallel there, without an upper bound. That's fine now, but something we could improve in the future

AlekSi (approved):

PR #382, opened 2023-03-23
Allow `jstest` runner to load files using a relative path

### LINKED ISSUE #295: Ensure that we can load files from within `mongo` shell

b1ron:
Closes #295.


b1ron:
I tested it with `array_comparison_correctness.js` and it is able to load the utils.js file. Good news.

internal/jstest/jstest.go
```diff
@@ -75,7 +73,10 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 			it := &item{
 				file: f,
 			}
-			it.out, it.err = runCommand(dir, "mongo", filepath.Join(volume, f))
+
+			// we set working_dir so a relative path is needed here
+			rel, _ := filepath.Rel("mongo", f)
```
b1ron, line 78: I purposefully ignored an error here as I cannot see under what circumstances an error would occur.

Should I return one anyway?

AlekSi (approved):

PR #374, opened 2023-03-20
Add diff tests for collection and database name

chilagrow:
Closes FerretDB/FerretDB#2210.

PR #383, opened 2023-03-23
Update Go driver configuration

b1ron:


AlekSi:
Diff tests will be fixed by #374

PR #384, opened 2023-03-24
Fix logs collection on failure

AlekSi:


chilagrow (approved):
Thanks for fixing the logs step and time out :hugs:

PR #396, opened 2023-03-31
Add runner for service tests

### LINKED ISSUE #318: Run service-tests with dance
https://github.com/mongodb-developer/service-tests

We don't want to use their Python code, but we could make a new dance runner that would use their test suites like https://github.com/mongodb-developer/service-tests/blob/master/post-5.0/test-suites/6.0/dbaas_core.yml

b1ron:
Closes #318.

This runner just calls another runner. The reason it's a separate runner is for CI and because it has a different configuration file.

Stats should be done before we add the runner to CI.


tests/dbaas_core.yml
```diff
@@ -0,0 +1,525 @@
...
+      - mongo/jstests/core/views/views_stats.js
+      - mongo/jstests/core/views/views_validation.js
+      - mongo/jstests/core/write/batch_write_collation_estsize.js
+      - mongo/jstests/core/write/collation_find_and_modify.js
+      - mongo/jstests/core/write/verify_update_mods.js
+      - mongo/jstests/core/write/write_commands_reject_unknown_fields.js
+      - mongo/jstests/core/write/write_result.js
+  mongodb:
+    stats:
+      # TODO: add stats.
```
b1ron, line 523: I still need to do this.

PR #407, opened 2023-04-13
Add diff tests for `scale` param in `dbStats`

rumyantseva:
See also https://github.com/FerretDB/FerretDB/pull/2418.

### DIFF
--- a/tests/diff/helpers.go
+++ b/tests/diff/helpers.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package diff contains diff tests.
 package diff
 
 import (
--- a/tests/diff.yml
+++ b/tests/diff.yml
@@ -5,8 +5,8 @@ args: ["-timeout=30s", "-shuffle=on", "-race", "./..."]
 results:
   ferretdb:
     stats:
-      expected_fail: 90
-      expected_pass: 37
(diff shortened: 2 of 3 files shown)


tests/diff/helpers.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package diff contains diff tests.
```
rumyantseva, line 15: Linter requires the description for the package.

chilagrow (approved):
LGTM

w84thesun (approved):

AlekSi:
diff tests fail for MongoDB. @rumyantseva @b1ron PTAL

b1ron (approved):
So we removed `MaxFloat` and it worked.

Why?

chilagrow (approved):

rumyantseva:
@chilagrow maxFloat +1 shouldn't cause an error either. There is an interesting way how `scale` works: 
```go
	case errors.Is(err, errLongExceededPositive):
		return math.MaxInt32, nil
```

I think I added the test for `MaxFloat` before I learned this interesting behavior.

rumyantseva:
So, basically, if `scale` is too big, error is not returned but `math.MaxInt32` is used instead.

PR #413, opened 2023-04-25
Improve service runner

### LINKED ISSUE #402: Improve service tests runner
#396 introduced a new runner, but with a small problem. The configuration file defines what runner should be used with what arguments. Those arguments were meant to be the general configuration parameters, not CLI flags (although that's true for all current runners). Therefore, the ExcludeArgs configuration option does not make much sense in that context.

Let's remove that option, implement handling of a leading `!` ourselves, and clarify that `jstest` and `service` runners support it (as well as other globs).

b1ron:
Closes #402.


AlekSi (changes requested):
Let's not use github.com/mattn/go-zglob and extend the scope of the issue.

b1ron:
The linter fails on [CI](https://github.com/FerretDB/dance/actions/runs/4870982774/jobs/8687388250?pr=413) but passes locally.

w84thesun (approved):
Nice one!

internal/jstest/jstest.go
```diff
@@ -29,28 +29,46 @@ import (
...
 	_ = ctx
 
 	filesM := make(map[string]struct{})
 
+	excludeFilesM := make(map[string]struct{})
+
 	// remove duplicates if globs match same files
 	for _, f := range args {
+		if strings.HasPrefix(f, "!") {
+			f = strings.TrimPrefix(f, "!")
```
AlekSi, line 43: There is a new `CutPrefix` that does both operation

AlekSi (approved):

internal/jstest/jstest.go
b1ron: Crisp!

AlekSi:
You do; that looks like a GitHub bug

AlekSi (approved):

PR #417, opened 2023-04-27
Allow uppercase Latin letters in database names

rumyantseva:
Closes https://github.com/FerretDB/FerretDB/issues/2451.

As uppercase letters are allowed, we should remove a diff test for it.

PR #418, opened 2023-05-01
Fix `mongo` configuration

b1ron:


b1ron:
@AlekSi we need to merge this first then #417 will fix the failing diff test.

PR #420, opened 2023-05-02
Fix Go driver config

b1ron:


b1ron:
@AlekSi we will have to ignore diff failures once more. IndexExamples fails as expected locally but I had to ignore for CI, unsure why.

PR #442, opened 2023-05-23
Add dotnet-example to CI and a new `command` runner

### LINKED ISSUE #433: Run `dotnet-example`
https://github.com/FerretDB/dotnet-example

It is critical for us to stay compatible with C# driver. Let's integrate that test into our dance tool and CI.

b1ron:
Closes #433.


b1ron:
I need to look into [this](https://github.com/FerretDB/dance/actions/runs/5058548343/jobs/9078810030?pr=442) failure.

AlekSi (dismissed):
Let's integrate it into the dance tool as a new simple command runner

internal/command/command.go
```diff
@@ -0,0 +1,81 @@
...
+func Run(ctx context.Context, dir string, args []string) (*internal.TestResults, error) {
+	// TODO https://github.com/FerretDB/dance/issues/20
+	_ = ctx
+
+	res := &internal.TestResults{
+		TestResults: make(map[string]internal.TestResult),
+	}
+
+	_ = dir
+	dir = args[len(args)-1]
```
AlekSi, line 38: Why we ignored passed `dir` and use the last argument instead?

internal/command/command.go
```diff
@@ -0,0 +1,81 @@
...
+
+	cmdName := args[0]
+	cmdArgs := args[1 : len(args)-1]
```
AlekSi, line 41:
```suggestion
	cmdName, cmdArgs := args[0], args[1:]
```

internal/command/command.go
```diff
@@ -0,0 +1,81 @@
...
+
+	for _, arg := range cmdArgs {
```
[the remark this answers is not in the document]
b1ron: You are right, this was my strange attempt installing dependencies but the .csproj file will handle this when `dotnet run` is ran I believe.

internal/command/command.go
b1ron: To fix the linter, it complaines that the argument dir is overwritten before first use.

AlekSi: Well, yeah, the linter is correct. Why we don't use passed dir?

w84thesun (dismissed):
LGTM!

b1ron:
@AlekSi PTAL when you get a chance.

AlekSi (approved):

PR #450, opened 2023-05-25
Add more jstests

b1ron:
They should cover features and bug fixes up to GA. I will do my best to catch everything and will create new issues as I go.


b1ron:
Closing for now.

PR #462, opened 2023-06-01
Fix `mongo` configuration

b1ron:


AlekSi (approved):

tests/mongo.yml
```diff
@@ -96,9 +92,6 @@ results:
       # https://jira.mongodb.org/browse/SERVER-14832
       - mongo/jstests/core/query/distinct/distinct_array1.js
 
-      # https://github.com/FerretDB/FerretDB/issues/2236
-      - mongo/jstests/core/query/distinct/distinct1.js
-
       # https://github.com/FerretDB/FerretDB/issues/2231
```
AlekSi, line 95: Hm, that one is closed. Please check

b1ron: Seems to be fixed in FerretDB/FerretDB/pull/2658. It also passes locally.

AlekSi: So should we remove it from here?

b1ron: Yes I believe so, it's fixed so I removed it from fail. Same deal with auth2.js. It was fixed in FerretDB/FerretDB/pull/2726 and it passes locally.

AlekSi: Oops! Let's remove them in a new PR

b1ron: Okie

PR #467, opened 2023-06-05
Add java-example to CI

### LINKED ISSUE #454: Simplify and run java-example
https://github.com/FerretDB/java-example

We should do the same as we did with https://github.com/FerretDB/dotnet-example: simplify and incorporate into the Dance tool. That way, we will ensure we stay compatible with the Java driver.

b1ron:
Closes #454.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,6 +2,10 @@
 	path = tests/dotnet-example
 	url = https://github.com/FerretDB/dotnet-example.git
 	branch = main
+[submodule "java-example"]
+	path = tests/java-example
+	url = https://github.com/FerretDB/java-example.git
+	branch = main
 [submodule "mongo"]
 	path = tests/mongo
 	url = https://github.com/mongodb/mongo.git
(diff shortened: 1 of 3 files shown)


b1ron:
I will look into this [failure](https://github.com/FerretDB/dance/actions/runs/5178162405/jobs/9329249470#step:7:95) on v1.3.0-dirty, could be a regression.

b1ron:
@AlekSi PTAL.

.gitmodules
```diff
@@ -15,3 +15,6 @@
 	path = dumps/mongodb-sample-databases
 	url = https://github.com/mcampo2/mongodb-sample-databases.git
 	branch = master
+[submodule "java-example"]
+	path = tests/java-example
+	url = https://github.com/FerretDB/java-example.git
```
AlekSi, line 20: Please add `branch = main` below (that will allow Dependabot to send PRs for that submodule) and move into the group above (that is sorted alphabetically)

tests/java-example.yml
```diff
@@ -0,0 +1,15 @@
+---
+runner: command
+dir: java-example/java
+args: [mvn, compile, exec:java, -Dexec.mainClass=com.start.Connection]
```
AlekSi, line 4: That does not actually _runs_ the example, only compiles it

b1ron: I believe that Maven plugin `exec:java` will build and then run it in the [same VM](https://www.digitalocean.com/community/tutorials/exec-maven-plugin-run-java-programs-maven-build).

See [here](https://github.com/FerretDB/dance/actions/runs/5178231009/jobs/9329409310#step:7:2262) for output from `listDatabases()`.

.gitmodules
b1ron: Done!

PR #473, opened 2023-06-06
Add all files for the dbaas_core suite

### LINKED ISSUE #424: Investigate all dbaas_core failures

b1ron:
Closes #424.


tests/dbaas_core.yml
```diff
@@ -156,27 +156,241 @@ args:
   - "!mongo/jstests/core/stage_debug/stages_or.js"
   - "!mongo/jstests/core/write/collection_truncate.js"
 
-  # include all tests in the following directories
-  - mongo/jstests/core/**/*.js
+  # include all the below core tests.
+  - mongo/jstests/core/administrative/*.js
+  - mongo/jstests/core/api/*.js
+  - mongo/jstests/core/benchrun_cmd_param_error.js
```
AlekSi, line 162: I think there should not be a need to list all core files: individual core files are excluded above, so that could be just `mongo/jstests/core/*.js`

b1ron: I do believe that will skip the files within sub-directories.

So, for example we'd end up skipping 73 files that test indexes:

```
➜  mongo git:(d853aab310c) ✗ ll jstests/core/index/*.js | wc -l
73
```

And there are many others:

```
  - mongo/jstests/core/query/*.js
  - mongo/jstests/core/query/all/*.js
  - mongo/jstests/core/query/and/*.js
  - mongo/jstests/core/query/array/*.js
  - mongo/jstests/core/query/count/*.js
  - mongo/jstests/core/query/cursor/*.js
  - mongo/jstests/core/query/date/*.js
  - mongo/jstests/core/query/dbref/*.js
  - mongo/jstests/core/query/distinct/*.js
  - mongo/jstests/core/query/elemmatch/*.js
 ```

**Comment by b1ron** [comment]
The numbers are likely still incorrect due to failures, among other issues that need investigation. We may also see discrepancies between the total number of tests ran as well.

#294 will need to fixed as well.

File: tests/dbaas_core.yml (hunk 1 above)
Comment (AlekSi) [reply tests/dbaas_core.yml:162]: Sure, we should list subdirectories. But why do that:

```
- mongo/jstests/core/benchrun_cmd_param_error.js
- mongo/jstests/core/benchrun_pipeline_updates.js
- mongo/jstests/core/bench_test1.js
- mongo/jstests/core/bench_test2.js
<...>
- mongo/jstests/core/client_metadata_ismaster.js
<...>
```

Those files are in the same directory

b1ron: Yes there's no reason to do that, you are correct.

I will fix that now.

AlekSi (approved):

PR #474, opened 2023-06-06
Set `TestData` object to allow more tests to run

### LINKED ISSUE #294: Investigate how we can set the `TestData` object
A lot of the jstests rely on a special object called `TestData` which holds test data. This is set via a .yml file as seen [here](https://github.com/mongodb/mongo/blob/master/buildscripts/resmokeconfig/suites/core.yml#L24) for the core suite and then passed to the `mongo` shell. It's also set [here](https://github.com/mongodb/mongo/blob/master/buildscripts/resmokelib/testing/testcases/jstest.py#L60-L65).

We should investigate how we can do this in dance as we enable more tests in #290.

b1ron:
Closes #294.

A naive attempt at solving #294 just to allow more tests to run.


internal/jstest/jstest.go
```diff
@@ -149,8 +150,17 @@ func runCommand(dir, command string, args ...string) ([]byte, error) {
 		return nil, err
 	}
 
-	args = append([]string{"--verbose", "--norc", "mongodb://host.docker.internal:27017/"}, args...)
+	// creates the TestData variable and sets the testName property for the shell
+	var eb strings.Builder
+	eb.WriteString("TestData = new Object();")
+	eb.WriteRune(' ')
+	f := filepath.Base(args[len(args)-1])
```
AlekSi, line 157: That looks sus. What is the resulting command-line?

internal/jstest/jstest.go
```diff
@@ -149,8 +150,17 @@ func runCommand(dir, command string, args ...string) ([]byte, error) {
...
+	testName := strings.TrimSuffix(f, filepath.Ext(f))
+	eb.WriteString(fmt.Sprintf("TestData.testName = '%s';", testName))
```
AlekSi, line 159: Please use `%q` and `fmt.Fprintf`

b1ron:
Those `uncaught exception: ReferenceError: can't access lexical declaration 'ret' before initialization :` errors are not related to this PR. They seem to related to files that don't invoke themselves as function. I tested this manually and on MongoDB version 6.0.6 and 6.0.3.

internal/jstest/jstest.go
b1ron: I also looked into setting this variable in a `.mongorc.js` file to avoid command-line arguments. This does not work because it only read this file after evaluating what has been passed on the [command line](https://www.mongodb.com/docs/mongodb-shell/mongoshrc/#.mongoshrc-configuration-file).

AlekSi: Sorry, but I think it is time to refactor that code. For example, `args` is always a single file, but instead of passing a single string, we continue treating `args` as a slice and modifying it in place, making our code needlessly complicated.

Please make the code in this file simpler.

AlekSi: And speaking of simplifying: `command` is always a `mongo` and not even a command – it is the Docker Compose service name. I think that function variable could be removed because it is not variable

b1ron:
@AlekSi PTAL

internal/jstest/jstest.go
```diff
@@ -141,20 +142,47 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
 	cmd.Dir = dir
 
 	log.Printf("Running %s", strings.Join(cmd.Args, " "))
 
 	return cmd.CombinedOutput()
 }
+
+// evalBuilder creates the TestData object and sets the testName property for the shell.
+// It optionally sets additional properties on the TestData object.
+func evalBuilder(script string, obj map[string]string) string {
```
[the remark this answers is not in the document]
AlekSi: Yeah, let's keep it for now

internal/jstest/jstest.go
```diff
@@ -141,20 +142,56 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
+	var eb bytes.Buffer
+	eb.WriteString("TestData = new Object();")
+	eb.WriteByte(' ')
```
AlekSi, line 172:
```go
fmt.Fprint(&eb, "TestData = new Object(); ")
```

I think we could simplify it a bit by using only one family of functions instead of both `fmt.FprintXXX` and `eb.WriteXXX`

internal/jstest/jstest.go
```diff
@@ -141,20 +142,56 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
+	scriptName := filepath.Base(script)
+	fmt.Fprintf(&eb, "TestData.testName = %q;", strings.TrimSuffix(scriptName, filepath.Ext(scriptName)))
+
+	if obj == nil {
+		return eb.String()
+	}
+
+	eb.WriteByte(' ')
+
+	// avoids extraneous space
```
AlekSi, line 182: or we could just not care to make code simpler :)

AlekSi (approved):

internal/jstest/jstest.go
```diff
@@ -141,20 +142,56 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
+	i := len(obj) - 1
+
+	for key, value := range obj {
+		fmt.Fprintf(&eb, "TestData.%s = %s;", key, value)
```
AlekSi, line 186:
```suggestion
		fmt.Fprintf(&eb, "TestData.%s = %q;", key, value)
```

?

internal/jstest/jstest.go
```diff
@@ -141,20 +142,56 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
+		"--verbose", "--norc", "mongodb://host.docker.internal:27017/",
+		"--eval", evalBuilder(script, nil), script,
+	}
+	dockerArgs = append(dockerArgs, shellArgs...)
+
 	cmd := exec.Command(bin, dockerArgs...)
 	cmd.Dir = dir
 
-	log.Printf("Running %s", strings.Join(cmd.Args, " "))
+	log.Printf("Running %s", script)
```
AlekSi, line 162: I think we could keep the old output

AlekSi (dismissed):

AlekSi (approved):

PR #476, opened 2023-06-07
Add and investigate all dbaas_core failures

### LINKED ISSUE #424: Investigate all dbaas_core failures

b1ron:
Closes #424.

### DIFF
--- a/tests/dbaas_core.yml
+++ b/tests/dbaas_core.yml
@@ -245,9 +245,6 @@ results:
       - mongo/jstests/core/api/api_params_getmore.js # https://github.com/FerretDB/FerretDB/issues/1733
       - mongo/jstests/core/api/api_version_create.js # https://github.com/FerretDB/FerretDB/issues/2813
       - mongo/jstests/core/api/api_version_find_and_modify.js # unknown field \"apiVersion\"
-      - mongo/jstests/core/api/api_version_new_50_language_features.js # https://github.com/FerretDB/FerretDB/issues/2590
-      - mongo/jstests/core/api/api_version_new_51_language_features.js # https://github.com/FerretDB/FerretDB/issues/2590
-      - mongo/jstests/core/api/api_version_new_52_language_features.js # https://github.com/FerretDB/FerretDB/issues/1467
       - mongo/jstests/core/api/api_version_pipeline_stages.js  # Stable API
       - mongo/jstests/core/api/api_version_unstable_fields.js  # Stable API
       - mongo/jstests/core/api/api_version_unstable_indexes.js # Stable API
@@ -261,6 +258,16 @@ results:
       - mongo/jstests/core/capped/cappeda.js # https://github.com/FerretDB/FerretDB/issues/2813
       - mongo/jstests/core/catalog/list_catalog.js # https://github.com/FerretDB/FerretDB/issues/176
       - mongo/jstests/core/catalog/list_collections1.js # https://github.com/FerretDB/FerretDB/issues/2815
+      # --- aggregation tests ---
+      # https://github.com/FerretDB/FerretDB/issues/2275
+      # https://github.com/FerretDB/FerretDB/issues/1469
+      # https://github.com/FerretDB/FerretDB/issues/414
+      # https://github.com/FerretDB/FerretDB/issues/1457
+      - mongo/jstests/core/command_let_variables.js
+      - mongo/jstests/core/command_let_variables_merge_only.js
+      - mongo/jstests/core/api/api_version_new_50_language_features.js # https://github.com/FerretDB/FerretDB/issues/2590
+      - mongo/jstests/core/api/api_version_new_51_language_features.js # https://github.com/FerretDB/FerretDB/issues/2590
+      - mongo/jstests/core/api/api_version_new_52_language_features.js # https://github.com/FerretDB/FerretDB/issues/1467
 
   mongodb:
     stats:
--- a/tests/dbaas_core.yml
+++ b/tests/dbaas_core.yml
@@ -239,6 +239,28 @@ results:
       - mongo/jstests/core/write/write_commands_reject_unknown_fields.js
     fail:
       # TODO: investigate failures.
+      - mongo/jstests/core/SERVER-23626.js # \x0D JavaScript code BSON type not implemented
+      - mongo/jstests/core/administrative/builtin_roles_external.js # https://github.com/FerretDB/FerretDB/issues/1536
+      - mongo/jstests/core/administrative/getlog2.js # \x0D JavaScript code BSON type not implemented
+      - mongo/jstests/core/api/api_params_getmore.js # https://github.com/FerretDB/FerretDB/issues/1733
+      - mongo/jstests/core/api/api_version_create.js # https://github.com/FerretDB/FerretDB/issues/2813
+      - mongo/jstests/core/api/api_version_find_and_modify.js # unknown field \"apiVersion\"
+      - mongo/jstests/core/api/api_version_new_50_language_features.js # https://github.com/FerretDB/FerretDB/issues/2590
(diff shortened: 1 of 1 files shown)


b1ron:
@AlekSi just FYI there's around 100 core tests that rely on feature flag server parameters.

```
> const params = db.adminCommand( { getParameter: "*" } );
> params.featureFlagCollModIndexUnique
{ "value" : true, "version" : "6.0" }
```

An example of it is [here](https://jira.mongodb.org/browse/SERVER-67846).

.github/workflows/dance.yml
```diff
@@ -3,7 +3,7 @@ name: Dance
 on:
   push:
     branches:
-      - main
+      - dbaas_core-failures
```
AlekSi, line 6: That's not needed because we test PRs bt the `pull_request` section below

.github/workflows/dance.yml
```diff
@@ -23,8 +23,9 @@ env:
 jobs:
   dance:
     name: dance
-    runs-on: ubuntu-22.04
-    timeout-minutes: 30
+    runs-on:
+      labels: ubuntu-20.04-16core
```
AlekSi, line 27: Yeah, that does not work

b1ron:
Can we please force merge to ignore the failing diff test as dbaas_core takes ~71m to finish.

.github/workflows/dance.yml
```diff
@@ -85,11 +86,20 @@ jobs:
       - name: Run init
         run: bin/task init
 
+      # The step that creates or updates the environment variable does not have access to the new value, 
+      # but all subsequent steps in a job will have access.
+      - name: Set GITHUB_ENV
+        run: |
+          if [[ ${{ matrix.test == 'dbaas_core' }} ]]; then
+            echo "FILE_CONCURRENCY=1" >> $GITHUB_ENV
```
b1ron, line 94: We do this because processing more than one file concurrently seems to make testing unstable for FerretDB.

I will create a separate issue later for this.

docker-compose.yml
```diff
@@ -39,6 +39,16 @@ services:
...
+    ulimits:
+      nofile:
+        soft: 64000
+        hard: 64000
+      nproc: 64000
+    deploy:
+      resources:
+        limits:
+          cpus: '2'
+          memory: 7G
```
b1ron, line 51: Matches the runner on CI ubuntu-22.04 and sets ulimits per the [recommendation](https://www.mongodb.com/docs/upcoming/reference/ulimit/).

tests/dbaas_core.yml
```diff
@@ -158,6 +158,10 @@ args:
   - "!mongo/jstests/core/stage_debug/stages_or.js"
   - "!mongo/jstests/core/write/collection_truncate.js"
 
+  # XXX culprit tests
```
rumyantseva, line 161: This comment is not super clear to me :)

rumyantseva (commented):
Great job figuring out the reasons for failed tests! I think we might need Alexey's review, but overall the PR looks good to me.

docker-compose.yml
chilagrow: Could you add this in a comment please. :hugs:

.github/workflows/dance.yml
chilagrow: Does this only happen on CI? I'm seeing this env var is set on CI but not for local runs.

internal/jstest/jstest.go
```diff
@@ -85,8 +85,14 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 		out  []byte
 	}
 
-	// tokens is a counting semaphore used to enforce a limit of 20 concurrent calls to runShellWithScript.
```
chilagrow, line 88: What is the new behaviour now? A comment would be useful here :hugs: 

If I understand correctly, on CI `FILE_CONCURRENCY` is set to 1, when is `token = 50` being used?

chilagrow (commented):
Few comments :hugs: great work detecting all failures

internal/jstest/jstest.go
b1ron: `$GITHUB_ENV` is just for CI and will not effect local runs. The above behavior is a slight refactor and an increase in the number of concurrent files being processed. This was to have tests complete faster locally.

.github/workflows/dance.yml
b1ron: FYI FILE_CONCURRENCY is set to 1 for all jobs due to https://github.com/actions/runner/issues/1126. It cannot be unset so all jobs will be slower overall that depend on the runner.

internal/jstest/jstest.go
chilagrow: Just a suggestion. I think we may want to improve naming of `FILE_CONCURRENCY` -> `tokens` -> `sema`. It's easy to get confused what is `FILE_CONCURRENCY` is :thinking:

chilagrow (commented):
Overall, looks good to me.

internal/jstest/jstest.go
b1ron: I changed it to `NUM_TOKENS`, not sure what else to call it in this context.

internal/jstest/jstest.go
```diff
@@ -85,8 +85,14 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
...
-	tokens := make(chan struct{}, 20)
+	tokens := 50
+
+	t := os.Getenv("NUM_TOKENS")
```
AlekSi, line 90: It is an internal library package, it should not read environment variables. Instead, we should use a flag in the main.go and pass it there

docker-compose.yml
AlekSi: I don't think we should _slow down_ our tests to match slow CI

b1ron: I agree, do you propose using a larger runner on CI?

internal/jstest/jstest.go
b1ron: Changed to `WORKERS` as the name implies its purpose.

b1ron:
:crossed_fingers:

docker-compose.yml
AlekSi: Maybe. But that's not related to this PR, I guess (and we probably should update the PR title).

internal/jstest/jstest.go
```diff
@@ -85,8 +83,7 @@ func Run(ctx context.Context, dir string, args []string) (*internal.TestResults,
 		out  []byte
 	}
 
-	// tokens is a counting semaphore used to enforce a limit of 20 concurrent calls to runShellWithScript.
-	tokens := make(chan struct{}, 20)
+	sema := make(chan struct{}, workers)
```
AlekSi, line 86: Now that we know the number of workers, I think we should change that code to launch that number of worker goroutines and send work items (files) to them via a channel, not launch one temporary goroutine per work item. That should simplify that code.

docker-compose.yml
b1ron: Cool!

internal/jstest/jstest.go
b1ron: But how do we ensure we process all the files if the number of workers is less than the number of files?

After we've finished iterating the workers there will still be files left in the channel to be processed, no?

b1ron: I think I know what to do.

b1ron: I think this pattern is more optimal because of long-lived goroutines and the avoidance of excessive concurrency.

AlekSi: The sender should close the working channel, workers should exit when the working channel is empty and closed, `sync.WaitGroup` should be used to wait for all of them to exit.

It is a bit more efficient (but only a little bit because goroutines are much more effective than, say, OS threads), but simplicity is the main reason – it is a very common and recognizable pattern in Go.

AlekSi (approved):

PR #508, opened 2023-07-10
Add auth examples to CI

### LINKED ISSUE #479: Run examples with and without authentication
We should run java-example, dotnet-example (and other examples like python-example once we have them) both with authentication enabled and disabled.

The easiest solution could be to pass the full MongoDB URI as a command-line argument instead of using hard-coded values:
* https://github.com/FerretDB/java-example/blob/6c92a3820d0ffd7afa1de8be4ca23433a8e1a677/java/src/main/java/com/start/Connection.java#L13
* https://github.com/FerretDB/dotnet-example/blob/4c18bfa1ba68ba4a9f70e2f795243d6aa32ee92f/Program.cs#L4

b1ron:
Closes #479.


b1ron:
I still need to add the Python example.

b1ron:
@AlekSi PTAL

AlekSi (commented):
I think submodules should be updated

b1ron:
@AlekSi do I need to add the python-example as a submodule in a separate PR?

tests/java-example-auth.yml
```diff
@@ -0,0 +1,19 @@
...
+
+results:
+  ferretdb:
+    stats:
+      expected_pass: 1
+
+  mongodb:
+    stats:
+      # PLAIN is used in MongoDB to perform LDAP authentication.
+      unexpected_fail: 1
```
AlekSi, line 19: `unexpected_fail` should not be expected :)

b1ron: :+1:

tests/dotnet-example-auth.yml
```diff
@@ -0,0 +1,21 @@
...
+args: 
+  - dotnet
+  - run
+  - mongodb://user:password@localhost:27017/?authMechanism=PLAIN
+
+results:
+  ferretdb:
+    stats:
+      # https://github.com/FerretDB/FerretDB/issues/2516
+      unexpected_fail: 1
```
AlekSi, line 13: That one too

.github/workflows/dance.yml
```diff
@@ -57,6 +61,12 @@ jobs:
         with:
           cache-key: dance
 
+      # set up Python build environment
+      - name: Install Python dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          pip install pymongo
```
AlekSi, line 68: I think `pip3 install pymongo` should work instead of two commands. Please check

.gitmodules
```diff
@@ -18,3 +18,6 @@
 	path = dumps/mongodb-sample-databases
 	url = https://github.com/mcampo2/mongodb-sample-databases.git
 	branch = master
+[submodule "tests/python-example"]
+	path = tests/python-example
+	url = https://github.com/FerretDB/python-example.git
```
AlekSi, line 23: Please use the same format as submodule above – with a different name and with `branch` (that is important for dependabot)

AlekSi (approved):

AlekSi (approved):

PR #501, opened 2023-07-05
Fix Go driver config

b1ron:

PR #509, opened 2023-07-12
Fix diff tests

b1ron:
We allow database names to start with numbers.


b1ron:
This will need to be merged before #501.

b1ron:
@AlekSi can we merge this so that #501 and #515 can also be merged.

AlekSi:
😛

PR #515, opened 2023-07-13
Bump deps

AlekSi:


.github/dependabot.yml
```diff
@@ -5,6 +5,10 @@ updates:
     directory: "/"
     labels: [deps]
     assignees: [AlekSi]
+    open-pull-requests-limit: 20
```
b1ron, line 8: Nice! I didn't know that was possible.

AlekSi: Ok, but please approve PR :)

b1ron (approved):

PR #541, opened 2023-08-09
Fix diff tests

b1ron:
It seems like we have changed some of the error messages for invalid database names and collection names.

### DIFF
--- a/tests/diff/names_test.go
+++ b/tests/diff/names_test.go
@@ -44,9 +44,9 @@ func TestDatabaseName(t *testing.T) {
 				expected := mongo.CommandError{
 					Name:    "InvalidNamespace",
 					Code:    73,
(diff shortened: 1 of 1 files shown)


chilagrow:
I feel this should be fixed on FerretDB side not on dance. Partly because the previous one seems more consistent, partly because we shouldn't break existing tests...

chilagrow:
If MongoDB returns `"errmsg" : "Invalid namespace: $cmd.f"` on failed database creation, isn't that what we used to have and we changed to something else :thinking:?

Anyway, this is just simple difference of error message. Let's just merge this PR for now :hugs:

chilagrow (approved):
:rocket:

noisersup (approved):

PR #548, opened 2023-08-16
Run Enmeshed tests on CI

### LINKED ISSUE #547: Run Enmeshed tests
See #537.

We should figure out passing secrets / environment variables.

We also should add some basic documentation about integrating other tests.

AlekSi:
Also bump deps, do small cleanups and port improvements from FerretDB.

Closes #547.


tests/enmeshed-runtime.yml
```diff
@@ -1,7 +1,6 @@
 ---
 runner: command
-dir: enmeshed-runtime
-args: ["sh", "-c", "npm i && CONNECTION_STRING=mongodb://localhost:27017 npx jest -i"]
+args: [../enmeshed-runtime.sh]
```
AlekSi, line 3: @b1ron I think we should document/promote that approach

b1ron: Nice, a much cleaner approach.

.github/settings.yml
```diff
@@ -0,0 +1,65 @@
...
+    description: Issues that are not ready to be worked on; PRs that should skip CI
+
+  # https://hacktoberfest.com/participation/#spam
+  - name: spam
+    color: "#C5DEF5"
+    description: Spam issues and PRs
+
+  - name: trust
+    color: "#00FF00"
+    description: PRs that can access Actions secrets
```
b1ron, line 65: The trust label was always a bit ambiguous to me, but now I know its intention.

b1ron: :+1:

PR #549, opened 2023-08-16
Add a guide for custom tests

### LINKED ISSUE #547: Run Enmeshed tests

b1ron:
Closes #547.


CONTRIBUTING.md
```diff
@@ -33,6 +33,59 @@ For example, see [mongo-go-driver tests configuration](https://github.com/Ferret
...
 It defines what test configuration to run; empty value runs all configurations.
 
+### Running custom tests
+
+In order to test your application with FerretDB you must use the `command` runner and add your repository as a submodule.
+The `command` runner will invoke any command and CLI arguments.
+
+For example if you want to test your Node.js application with FerretDB, you would do the following:
+
+1. Add the submodule to dance `git submodule add https://github.com/my-app`
```
AlekSi, line 43:
```suggestion
1. Add the submodule to dance `git submodule add https://github.com/my-org/my-app.git`.
```

CONTRIBUTING.md
```diff
@@ -33,6 +33,59 @@ For example, see [mongo-go-driver tests configuration](https://github.com/Ferret
...
+2. Create a shell script in the `tests` directory called `my-app-runtime.sh` with the required logic needed for your test
```
AlekSi, line 44:
```suggestion
2. Create a shell script in the `tests` directory called `my-app.sh` with the required logic needed for your test.
```

CONTRIBUTING.md
```diff
@@ -33,6 +33,59 @@ For example, see [mongo-go-driver tests configuration](https://github.com/Ferret
...
+3. Create a YAML file called `my-app.yml` in the `tests` directory and provide the `args` field with the shell script so that the runner can invoke it
```
AlekSi, line 45:
```suggestion
3. Create a YAML file called `my-app.yml` in the `tests` directory and provide the `args` field with the shell script so that the runner can invoke it.
```

CONTRIBUTING.md
```diff
@@ -33,6 +33,59 @@ For example, see [mongo-go-driver tests configuration](https://github.com/Ferret
...
+4. Start the environment `bin/task env-up DB=ferretdb`
```
AlekSi, line 46:
```suggestion
4. Start the environment `bin/task env-up DB=ferretdb`.
```

CONTRIBUTING.md
```diff
@@ -33,6 +33,59 @@ For example, see [mongo-go-driver tests configuration](https://github.com/Ferret
...
+5. Run the test `bin/task dance DB=ferretdb TEST=my-app`
```
AlekSi, line 47:
```suggestion
5. Run the test `bin/task dance DB=ferretdb TEST=my-app`.
```

CONTRIBUTING.md
```diff
@@ -33,6 +33,59 @@ For example, see [mongo-go-driver tests configuration](https://github.com/Ferret
...
+
+#### Shell script
```
AlekSi, line 49: I think we should point to existing example instead of having another copy of it there

CONTRIBUTING.md
AlekSi: Please also note that this is documented below. So the placement of this section might not be the best.

CONTRIBUTING.md
```diff
@@ -33,6 +33,59 @@ For example, see [mongo-go-driver tests configuration](https://github.com/Ferret
 `TEST` environment variable should have the value `mongo-go-driver`, or be empty.
 It defines what test configuration to run; empty value runs all configurations.
 
+### Running custom tests
```
AlekSi, line 36:
```suggestion
### Adding tests
```

We want to document how to add tests to Dance CI, not just how to run them with Dance (which does not have much value – they already know how to run their tests, and using Dance has no benefits for them)

CONTRIBUTING.md
b1ron: And the YAML too?

CONTRIBUTING.md
b1ron: Fixed

CONTRIBUTING.md
AlekSi: Why not? It also exist and we also don't want to make another copy of it

b1ron: Done, I used the java-example as a reference

AlekSi: I'm unsure what is "done", but the example is still in the text. We point to existing examples in addition to providing one there, not _instead_. And we should not have them in two places.

b1ron: I see, so just link to examples and not remove all the text related to examples?

b1ron: @AlekSi PTAL.

AlekSi (approved):

PR #562, opened 2023-08-21
Split the dbaas_core config to run in parallel

### LINKED ISSUE #424: Investigate all dbaas_core failures

b1ron:
Closes #424.

~Some glob file patterns match more files that others so it's not guaranteed that work will be evenly distributed.~

I avoided globs and split the files into segments so that work will be distributed evenly in terms of the number of the files.


b1ron:
We are seeing a reduction in completion time by 56% due to this change. We've basically doubled our performance on CI. Hopefully, we also see consistent results from hereon.

tests/dbaas_core-0.yml
AlekSi, line 1: Is there a better way to split tests / name files? By some common theme or something?

b1ron: Isn't it better to have it randomized so that we distribute load somewhat evenly to avoid any one VM taking most of the load?

Some of the directories contain more files but do less work vice versa. I honestly don't know of a good approach to split the files in any other way without continuous trial and error.

AlekSi: Ok

PR #570, opened 2023-08-25
Split config packages

### LINKED ISSUE #563: Split configuration loading into two package
Currently, dance has structures and functions that load and parse YAML configuration files and transform them into a form used by other packages all in one package. Some `struct`s have two representations (`Results` and `ResultsYAML`), while others have only one used in both contexts (`Stats`). That makes it harder to introduce new features (#463) and already lead to one abandoned PR (#327).

We should split this code into two packages: `internal/config` and `internal/configload`. The first one should be used by other packages and "business logic", while the second should export a single `Load` function that would return `internal/config.Config` `struct`.

Do not do other features or refactorings in this issue.

b1ron:
Closes #563.


internal/config/config.go
```diff
@@ -23,30 +25,84 @@ import (
...
+	RunnerTypeYCSB RunnerType = "ycsb"
+)
+
+// RunnerType represents the type of test runner used in the configuration.
+type RunnerType string
+
+// Stats represent the expected/actual amount of
+// failed, skipped and passed tests.
+type Stats struct {
+	UnexpectedRest int `yaml:"unexpected_rest"`
```
AlekSi, line 48: This package should not have structs with YAML tags

internal/configload/configload.go
```diff
@@ -12,113 +12,131 @@
...
 
-// ConfigYAML is a yaml tests representation of the Config struct.
-//
-// It is used only to fetch data from file. To get any of
-// the dance configuration data it should be converted to
-// Config struct with Convert() function.
+// ConfigYAML represents the YAML-based configuration for the testing framework.
 //
 //nolint:govet // we don't care about alignment there
 type ConfigYAML struct {
```
AlekSi, line 31: Those types should not be exported

internal/configload/configload.go
```diff
@@ -0,0 +1,221 @@
...
+	"golang.org/x/exp/maps"
+	"gopkg.in/yaml.v3"
+
+	"github.com/FerretDB/dance/internal/config"
+)
+
+// configYAML represents the YAML-based configuration for the testing framework.
+//
+//nolint:govet // we don't care about alignment there
+type configYAML struct {
```
b1ron, line 31: I can remove the YAML suffix if needed.

internal/configload/configload.go
```diff
@@ -0,0 +1,279 @@
...
+		if !validStatus(r.Default) {
+			return fmt.Errorf("invalid default result: %q", *origDefault)
+		}
+
+		if *commonDefault != "" && r.Default != "" {
+			return errors.New("default value cannot be set in common, when it's set in database")
+		}
+
+		// XXX this doesn't work
+		r.Default = *commonDefault
```
b1ron, line 275: We can't set the default value here, so we have to set it [here](https://github.com/FerretDB/dance/pull/570/files#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR181) instead which makes this method incomplete.

internal/configload/configload.go
```diff
@@ -0,0 +1,284 @@
...
+		origDefault = commonDefault
```
b1ron, line 280: We can't set the default value here, so we have to set it [here](https://github.com/FerretDB/dance/pull/570/files#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR181) instead which makes this method incomplete.

internal/config/config_test.go
```diff
@@ -12,178 +12,217 @@
...
 				},
-				MongoDB: &TestsConfig{
+				MongoDB: &TestConfig{
 					Default: "pass",
 				},
 			},
 		},
-		"FillAndValidateDefaultDuplicate": {
+		// XXX fails because this error is now returned in fillAndValidate
+		"MergeTestConfigsDefaultDuplicate": {
```
b1ron, line 187: I need to write a new test for this.

internal/configload/configload.go
```diff
@@ -0,0 +1,285 @@
...
+
+	validStatus := func(status ic.Status) bool {
+		s := ic.Status(strings.ToLower(string(status)))
+		_, ok := ic.KnownStatuses[s]
+
+		return ok
+	}
+
+	if c.Results.Common.Default == nil {
+		c.Results.Common.Default = new(ic.Status)
```
b1ron, line 248: `ic.Status` is not a composite type so I cannot allocate it using that syntax, right?

internal/config/config.go
```diff
@@ -23,42 +25,195 @@ import (
...
+	Runner RunnerType
+	Dir    string
+	// Args contains additional arguments for the test runner.
+	Args    []string
+	Results Results
+}
+
+// Results stores the expected test results for different databases.
+type Results struct {
+	Common   *TestConfig
```
AlekSi, line 72: Let's remove that field. The idea that `configload` will do all needed transformations, merges, etc, and then the rest of the code could work with a simple `Config` type by using `Config.FerretDB` or `Config.For("ferretdb")`

b1ron: Do you mean here

internal/config/config.go
```diff
@@ -23,42 +24,137 @@ import (
...
+
+	// RunnerTypeYCSB indicates a YCSB test runner.
+	RunnerTypeYCSB RunnerType = "ycsb"
+)
+
+// RunnerType represents the type of test runner used in the configuration.
+type RunnerType string
+
+// Stats represent the expected/actual amount of
+// failed, skipped and passed tests.
```
AlekSi, line 45: FWIW, I think we should follow the same order of fields as in this comment: failed, skipped, passed. Think of it like a traffic lights: 🔴, 🟡, 🟢.

internal/config/config.go
```diff
@@ -279,3 +381,13 @@ func (tc *TestsConfig) toMap() (map[string]status, error) {
 
 	return res, nil
 }
+
+// CheckDuplicates checks for duplicate test names or prefixes within a *TestConfig instance.
+// It returns an error if any duplicates are found.
+func (tc *TestConfig) CheckDuplicates() error {
```
AlekSi, line 387: That should be in the `configload` package. Once the config is loaded, it is valid, does not contain duplicates, etc

internal/configload/configload.go
b1ron: @AlekSi is there a directive I can use to ignore this in the linter?

AlekSi (commented):
CI fails. Please run linters locally

b1ron:
> CI fails. Please run linters locally

I did and fixed what could be fixed. The other failure is related to a syntax error, see [here](https://github.com/FerretDB/dance/actions/runs/6115959796/job/16600415356?pr=570#step:4:104). I mentioned it in [this](https://github.com/FerretDB/dance/pull/570#discussion_r1315143899) comment as well.

internal/configload/configload.go
```diff
@@ -249,8 +250,7 @@ func (c *config) fillAndValidate() error {
 
 	// initialize default common field if it's nil
 	if c.Results.Common.Default == nil {
-		c.Results.Common.Default = new(ic.Status)
-		*c.Results.Common.Default = ic.Pass
+		c.Results.Common.Default = pointer.To(ic.Pass)
```
b1ron, line 253: Oh wow!

AlekSi (approved):

PR #584, opened 2023-09-11
SQLite integration

### LINKED ISSUE #463: Integrate FerretDB testing with SQLite
Let's start by thinking (or completely rethinking) the YAML structure. Could we use YAML anchors or something? or just some common section?

b1ron:
Closes #463.


internal/configload/configload.go
```diff
@@ -36,20 +36,23 @@ type config struct {
...
-	Skip    []any      `yaml:"skip"`
-	Pass    []any      `yaml:"pass"`
-	Ignore  []any      `yaml:"ignore"`
+	Default     *ic.Status `yaml:"default"`
+	Stats       *stats     `yaml:"stats"`
+	Fail        []any      `yaml:"fail"`
+	Skip        []any      `yaml:"skip"`
+	Pass        []any      `yaml:"pass"`
+	Ignore      []any      `yaml:"ignore"`
+	IncludeFail []any      `yaml:"include_fail"`
```
AlekSi, line 55: I wonder if all those can and should be `[]string`

internal/config/config.go
```diff
@@ -70,6 +70,7 @@ type Config struct {
 // Results stores the expected test results for different databases.
 type Results struct {
 	FerretDB *TestConfig
```
AlekSi, line 72: Let's rename that to `PostgreSQL`

internal/configload/configload.go
```diff
@@ -36,20 +36,23 @@ type config struct {
 	Dir     string        `yaml:"dir"`
 	Args    []string      `yaml:"args"`
 	Results struct {
-		Common   *testConfig `yaml:"common"`
-		FerretDB *testConfig `yaml:"ferretdb"`
-		MongoDB  *testConfig `yaml:"mongodb"`
+		Includes map[string][]string `yaml:"includes"`
+		Common   *testConfig         `yaml:"common"`
+		FerretDB *testConfig         `yaml:"ferretdb"`
```
AlekSi, line 41:
```suggestion
		PostgreSQL *testConfig         `yaml:"ferretdb"`
```

Preserve the YAML tag for compatibility for now. We will update it later everywhere.

internal/configload/configload.go
```diff
@@ -36,20 +36,23 @@ type config struct {
 	Dir     string        `yaml:"dir"`
 	Args    []string      `yaml:"args"`
 	Results struct {
-		Common   *testConfig `yaml:"common"`
-		FerretDB *testConfig `yaml:"ferretdb"`
-		MongoDB  *testConfig `yaml:"mongodb"`
+		Includes map[string][]string `yaml:"includes"`
+		Common   *testConfig         `yaml:"common"`
```
AlekSi, line 40: Let's create a separate issue for removing that and add `// TODO <url>` comment

internal/configload/configload.go
b1ron: We cannot do that unfortunately as we use nested mappings for regex patterns.

b1ron: Unless I just use the same approach and unmarshal them into a field called `IncludeRegexp`. I will see.

b1ron: I added an `IncludeSkip` field to simplify how we handle regex patterns. But [diff.yml](https://github.com/FerretDB/dance/blob/main/tests/diff.yml) now breaks as we do not use type assertion so we cannot have nested mappings.

I propose we attempt to change the diff config in the near future instead. @AlekSi what do you think?

AlekSi (changes requested):
Runs fail hard: https://github.com/FerretDB/dance/actions/runs/6271907108/job/17032412184?pr=584
They should not

b1ron:
> Runs fail hard: https://github.com/FerretDB/dance/actions/runs/6271907108/job/17032412184?pr=584
They should not

@AlekSi Fixed.

AlekSi (approved):

PR #597, opened 2023-09-22
Fix and reorganize all config files

### LINKED ISSUE #596: Fix all config files for SQLite
Because of #584 all config files need to be fixed, and reorganized if necessary.

b1ron:
Closes #596.


b1ron:
@AlekSi CI seems to be in a somewhat broken state with 13 expected checks for `ferretdb`.

internal/configload/configload.go
```diff
@@ -37,7 +37,7 @@ type config struct {
 		// Includes is a mapping that allows us to merge sequences together,
 		// which is currently not possible in the YAML spec - https://github.com/yaml/yaml/issues/48
 		Includes   map[string][]string `yaml:"includes"`
-		PostgreSQL *testConfig         `yaml:"ferretdb"` // TODO preserving YAML tag for compatibility, will update later
+		PostgreSQL *testConfig         `yaml:"postgresql"`
```
AlekSi, line 40: Please move code all code changes into a separate PR

b1ron:
workload A on SQLite is an order of magnitude slower than the read-only workload B, this will be helpful to measure performance.

AlekSi (approved):

PR #611, opened 2023-10-05
Refactor more

b1ron:


internal/gotest/gotest.go
```diff
@@ -102,10 +108,8 @@ func Run(ctx context.Context, dir string, args []string, verbose bool, parallel
 		case actionSkip:
 			result.Status = config.Skip
 		case actionBench, actionCont, actionOutput, actionPause, actionRun:
-			fallthrough
-		default:
-			result.Status = config.Unknown
```
AlekSi, line 107: I just realized that his file was completely correct: the test run must end with one of pass/skip/fail. If any other event is received, the test is in an unknown state.

Let's revert those changes

internal/configload/configload_test.go
```diff
@@ -142,14 +93,14 @@ func TestIncludes(t *testing.T) {
...
+					Names: []string{"x", "a", "b", "c"},
 				},
 			},
 			expectedErr: nil,
 		},
 		"IncludeIgnore": {
-			in: &testConfig{
-				Default:       (*ic.Status)(pointer.ToString("ignore")),
+			in: &backend{
+				Default:       ic.Status("ignore"),
```
AlekSi, line 103: That's `ic.Ignore`. The same in other places.

internal/configload/configload_test.go
```diff
@@ -158,7 +109,7 @@ func TestIncludes(t *testing.T) {
 			},
 			expected: &ic.TestConfig{
 				Ignore: ic.Tests{
-					Names: []string{"a", "b", "c", "i"},
+					Names: []string{"i", "a", "b", "c"},
```
AlekSi, line 112: We probably should sort test names in backend.go

AlekSi (approved):

PR #614, opened 2023-10-06
Fix all config files

b1ron:
I manually added the full paths to tests that should be skipped.

And `TestUnifiedSpecs` is now ignored by MongoDB too, as it seems to contain flaky tests.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -193,10 +193,10 @@ func (tc *TestConfig) Compare(results *TestResults) (*CompareResult, error) {
 			switch testRes.Status {
 			case Fail:
 				compareResult.ExpectedFail[test] = testResOutput
-			case Pass:
-				compareResult.ExpectedPass[test] = testResOutput
 			case Skip:
 				compareResult.ExpectedSkip[test] = testResOutput
+			case Pass:
+				compareResult.ExpectedPass[test] = testResOutput
 			case Ignore:
 				fallthrough
 			case Unknown:
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -50,9 +50,9 @@ type Stats struct {
 	UnexpectedSkip int
 	UnexpectedPass int
 	UnexpectedRest int
-	Fail           int
-	Skip           int
-	Pass           int
+	ExpectedFail   int
+	ExpectedSkip   int
+	ExpectedPass   int
 }
 
 // Config represents the configuration settings for the test execution.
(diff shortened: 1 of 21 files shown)


internal/config/config.go
```diff
@@ -50,9 +50,9 @@ type Stats struct {
 	UnexpectedSkip int
 	UnexpectedPass int
 	UnexpectedRest int
-	ExpectedFail   int
-	ExpectedSkip   int
-	ExpectedPass   int
+	Fail           int
+	Skip           int
+	Pass           int
```
AlekSi, line 55: No, those fields should stay as is

internal/configload/stats.go
```diff
@@ -18,9 +18,12 @@ import ic "github.com/FerretDB/dance/internal/config"
 
 // stats represents the YAML representation of expected stats.
 type stats struct {
-	Fail int `yaml:"fail"`
-	Skip int `yaml:"skip"`
-	Pass int `yaml:"pass"`
+	UnexpectedFail int `yaml:"unexpected_fail"`
+	UnexpectedSkip int `yaml:"unexpected_skip"`
+	UnexpectedPass int `yaml:"unexpected_pass"`
```
AlekSi, line 23: We are going back and forth there. Those fields should never be present in the YAML configuration file – we don't expect the unexpected. `fail`, `skip`, and `pass` are expected values. Unexpected values should not be in the file.

We split `config` and `configload` into separate packages exactly to have them separate. Why are we trying to blend them together again?

AlekSi (changes requested):

internal/configload/stats.go
b1ron: My bad for that confusion you are completely right. It's a temporary fix for when we removed our support for regex, as that logic accounted for Go tests that should be skipped. So, now they will be marked as `UnexpectedSkip`. All those other unexpected fields are redundant so I will remove them. I will also add a comment.

AlekSi: Let's remove `UnexpectedSkip` and all associated tests. Adding a comment does not change the fact that we expect unexpected things

b1ron: OK cool.

b1ron: I tried to remove all associated tests with unexpected skips but it didn't work as some tests rely on variables and helpers, etc. that are contained within other files.

I think I have a less tedious approach now.

b1ron:
@AlekSi interestingly, workloadc the [READs](https://github.com/FerretDB/dance/actions/runs/6485067358/job/17610360564?pr=614#step:7:85) seem to increase linearly in time by roughly 10 seconds per read. I will look into more.

AlekSi (approved):

PR #624, opened 2023-10-12
Add capped collection tests

### LINKED ISSUE #604: Enable all tests for capped collections
We should enable all tests for capped collections we can find. As much as possible should pass when run against MongoDB.

b1ron:
Closes #604.


b1ron:
@AlekSi I had to remove some tests that do not assert on [create](https://github.com/mongodb/mongo/blob/e7dc7c9e239c1890caa679a7f4fd8af79270155c/jstests/core/capped/capped.js#L10) because they don't return a non-zero exit code, when they should actually fail.

If we parse the output for errors then we can avoid this in the future, the shell's output is a mix of plain text and JSON.

AlekSi (approved):

Issue #248, opened 2022-11-18
Run Meteor tests


AlekSi:
https://github.com/meteor/meteor/blob/devel/DEVELOPMENT.md

Most of them will fail without OpLog. That's expected.

PR #634, opened 2023-10-24
Add oplog meteor integration tests

### LINKED ISSUE #633: Add Meteor integration tests to CI
How to handle subtests?

b1ron:
Closes #633, #248.

This PR also enables the MongoDB service as a single node replica set, as well as a bunch of other clean up related stuff and general fixes.


tests/mongo.yml
```diff
@@ -58,7 +58,6 @@ args:
   - mongo/jstests/noPassthroughWithMongod/capped4.js
   - mongo/jstests/aggregation/sources/collStats/query_exec_stats.js
   - mongo/jstests/core/notablescan_capped.js
-  - mongo/jstests/core/local_tail_capped.js
```
b1ron, line 61: It keeps failing on CI against MongoDB so I removed it.

tests/meteor.sh
```diff
@@ -0,0 +1,18 @@
...
+export METEOR_LOCAL_DIR=~/.meteor
+
+export MONGO_URL="mongodb://localhost:27017/"
+
+export MONGO_OPLOG_URL="mongodb://localhost:27017/local?replicaSet=rs0&directConnection=true"
+
+TINYTEST_FILTER="mongo-livedata - doc fetcher" ./packages/test-in-console/run.sh --once
+TINYTEST_FILTER="mongo-livedata - oplog - cursorSupported" ./packages/test-in-console/run.sh --once
+TINYTEST_FILTER="mongo-livedata - oplog - entry skipping" ./packages/test-in-console/run.sh --once
+TINYTEST_FILTER="mongo-livedata - oplog - x - implicit collection creation" ./packages/test-in-console/run.sh --once
```
b1ron, line 18: All these tiny tests get run in the same script so currently we do not have details on each specific sub-test's outcome. I will figure out a better way to pass them to the command runner in the future.

b1ron:
It's still a bit broken see [here](https://github.com/FerretDB/dance/actions/runs/6723565670/job/18273923767?pr=634#step:7:723). I think I simply need to adjust the failures to relative the paths.

internal/command/command.go
```diff
@@ -53,3 +65,31 @@ func Run(ctx context.Context, dir string, args []string) (*config.TestResults, e
 
 	return res, nil
 }
+
+func runAllAsCommands(ctx context.Context, dir string, files []string) (*config.TestResults, error) {
```
chilagrow, line 69: suggest something more specific like `runExecutableFiles`, let's also add godoc. :hugs:

tests/meteor.yml
```diff
@@ -0,0 +1,31 @@
...
+      fail: 4
+    fail:
+      - meteor-doc-fetcher.sh
+      - meteor-oplog-cursor-supported.sh
+      - meteor-oplog-entry-skipping.sh
+      - meteor-oplog-x-implicit-collection-creation.sh
+
+  mongodb:
+    stats:
+      pass: 4
```
b1ron, line 31: These are all failures at the moment because I need to configure the MongoDB service a single node replica set.

internal/command/command.go
```diff
@@ -29,6 +29,18 @@ import (
...
 func Run(ctx context.Context, dir string, args []string) (*config.TestResults, error) {
+	allCommands := true && len(args) > 1
+
+	for _, arg := range args {
+		if !strings.HasSuffix(arg, ".sh") {
+			allCommands = false
+		}
+	}
+
+	if allCommands {
```
chilagrow, line 40:
```suggestion
	if !slices.ContainsFunc(args, func(arg string) bool { return !strings.HasSuffix(arg, ".sh") }) {
```

Do I understand correctly that if there is any arg that does not contain `.sh` suffix, we shouldn't run each file?

You probably want to update function godoc, highlighting special case with .sh handling.

tests/meteor-doc-fetcher.sh
```diff
@@ -0,0 +1,16 @@
...
+
+npm install -g npm@latest
+
+export PUPPETEER_DOWNLOAD_PATH=~/.npm/chromium
+
+export METEOR_LOCAL_DIR=~/.meteor
+
+export MONGO_URL="mongodb://localhost:27017/"
+
+# https://github.com/FerretDB/FerretDB/blob/main/cmd/envtool/envtool.go#L155
```
chilagrow, line 13: I don't understand the purpose of this link. What does this mean? Should this be issue link instead of a link to line number which can change?

tests/meteor.yml
```diff
@@ -0,0 +1,31 @@
+---
+runner: command
+dir: meteor
+args:
```
chilagrow, line 4: I'm not sure why we don't call each file individually rather than calling 4 of them at once. It seems simpler that way because less change in the go code. Perhaps the way implemented is how it should be done in dance to report result? :thinking:

chilagrow (commented):
Some comments, but keep in mind that I'm not familiar with dance setup, so all my comments are suggestions and questions rather than change request :hugs:

tests/meteor-doc-fetcher.sh
b1ron: It was supposed to show that we should keep the replica set name consistent in all places. It's mongodb-rs in envtool.go, I need to mention that properly.

internal/command/command.go
b1ron: Yes, your understanding is correct. I will add comments too.

tests/meteor.yml
b1ron: But then we'd have X individual YAML files where X is the number of subtests and our CI would have a lot more jobs. And I suspect that we'll be adding a lot more of these in the future.

(mongodb, meteor-doc-fetcher)
(mongodb, meteor-oplog-foo)
(mongodb, meteor-oplog-bar)
(mongodb, meteor-oplog-baz)
(mongodb, meteor-oplog-x)
(mongodb, meteor-oplog-y)
(mongodb, meteor-oplog-z)
.
.
.

b1ron: > Perhaps the way implemented is how it should be done in dance to report result?

Perhaps you're right. Let me make the other changes first and then we can decide.

tests/meteor-doc-fetcher.sh
b1ron: I removed the link to avoid confusion.

tests/meteor.yml
b1ron: Not sure how we can spin up a different environment per DB.

```
      - name: Start environment
        run: bin/task env-up-detach DB=${{ matrix.db }}
        env:
          FERRETDB_IMAGE: ghcr.io/ferretdb/ferretdb-dev:main
      
      - name: Start MongoDB environment
        if: ${{ matrix.db == 'mongodb' }}
        run: bin/task env-up-repl
```

Taskfile.yaml
```diff
@@ -41,24 +40,15 @@ tasks:
       - >
         docker compose up --always-recreate-deps --force-recreate --remove-orphans --renew-anon-volumes -t 0 --detach
         --build {{.DB}}
+      - task: init
```
b1ron, line 43: Fails here because we need a stub.

```
MongoServerError: no such command: 'replSetInitiate'
exit status 1
```

tests/meteor.yml
b1ron: Actually, it's a good idea to group subtests and run them in a separate job to reduce the overall completion time. For now I think it's fine but as is, but in the future we can scale it this way.

.gitmodules
```diff
@@ -26,3 +26,6 @@
 	path = tests/python-example
 	url = https://github.com/FerretDB/python-example.git
 	branch = main
+[submodule "meteor"]
+	path = tests/meteor
+	url = https://github.com/meteor/meteor.git
```
AlekSi, line 31: `branch`

tests/meteor.yml
```diff
@@ -0,0 +1,31 @@
...
+  - ../meteor-doc-fetcher.sh
+  - ../meteor-oplog-cursor-supported.sh
+  - ../meteor-oplog-entry-skipping.sh
+  - ../meteor-oplog-x-implicit-collection-creation.sh
+
+results:
+  postgresql:
+    stats:
+      fail: 4
```
AlekSi, line 13: Issue links

internal/command/command.go
```diff
@@ -28,7 +29,17 @@ import (
 // Run runs generic test.
 // It runs a command with arguments in a directory and returns the combined output as is.
 // If the command exits with a non-zero exit code, the test fails.
+// If all args are executable files with .sh extension it runs each file in a separate CommandContext.
 func Run(ctx context.Context, dir string, args []string) (*config.TestResults, error) {
+	allExecutableFiles := !slices.ContainsFunc(args, func(arg string) bool {
```
AlekSi, line 34: No, we are not doing that. `command` handler should run one command. We are not running one or multiple commands based on command arguments.

Please create separate YAML configurations or find some other solution.

AlekSi (changes requested):

b1ron:
re SQLITE_BUSY error, [ignoring them all](https://github.com/FerretDB/dance/pull/634/commits/d87a4194a8a082ac4b2fabf1f065a41ba9f36acb) for now as test names vary.

b1ron:
The only thing left to fix is the config for **(mongodb, mongo-go-driver)**.

AlekSi (commented):
There is a merge conflict now

tests/restheart-auth.yml
```diff
@@ -1,7 +1,7 @@
 ---
 runner: command
 dir: restheart
-args: [../restheart.sh]
+args: [../restheart-auth.sh]
```
b1ron, line 4: It was using the non-auth script before.

Taskfile.yaml
```diff
@@ -116,6 +106,16 @@ tasks:
         docker compose run --rm mongosh mongosh mongodb://host.docker.internal:27017/
         --verbose --eval 'disableTelemetry()' --shell
 
+  init-repl:
+    desc: "Initializes a new replica set"
+    cmds:
+      - >
+        docker compose run --rm mongosh mongosh
+        'mongodb://host.docker.internal:27017/'
+        --eval 'db.getSiblingDB("admin").runCommand({replSetInitiate: {_id: "mongodb-rs", members:[{_id: 0, host: "localhost"}]}});'
```
AlekSi, line 115: Can we use `rs.initiate` as we do in FerretDB? See https://github.com/FerretDB/FerretDB/blob/2664b9a5dd570d8157fc690a018934171908575c/cmd/envtool/envtool.go#L184

docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
     build:
       context: ./build/deps
       dockerfile: mongo6.Dockerfile
-    command: --setParameter enableTestCommands=1
+    command: --config /etc/mongod.conf --setParameter enableTestCommands=1
```
AlekSi, line 51: While we are on it, could we move `enableTestCommands=1` to the config file too?

build/mongod.conf
```diff
@@ -0,0 +1,16 @@
...
+    mode: disabled
+
+security:
+  authorization: disabled
+
+replication:
+   oplogSizeMB: 512
+   replSetName: mongodb-rs
+
+# Run mongod in a quiet mode that attempts to limit the amount of output.
```
AlekSi, line 14: I think that's kind of obvious :)

tests/meteor-doc-fetcher.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+set -ex
+
+npm install -g npm@latest
```
AlekSi, line 5: I think we should assume that node.js and npm are already recent enough

Taskfile.yaml
b1ron: I needed to get a connection to the admin DB, but it now seems to work using the shell method. Strange.

AlekSi (approved):

PR #658, opened 2023-11-10
Bump Go and tools deps

AlekSi:


CONTRIBUTING.md
```diff
@@ -1,8 +1,10 @@
 # Contributing
 
 The dance tool and tests run on the host; macOS, Linux, and Windows are expected to work.
-Databases under test (FerretDB and MongoDB) may be running on the host or inside Docker; Docker Compose configuration is provided for convenience but not required.
-In particular, the FerretDB development cycle (fix-compile-run-dance) is faster with it running on the host as it does not involve Docker image building or PostgreSQL restarts.
+Databases under test (FerretDB and MongoDB) may be running on the host or inside Docker;
+Docker Compose configuration is provided for convenience but not required.
+In particular, the FerretDB development cycle (fix-compile-run-dance) is faster with it running on the host
```
b1ron, line 6:
```suggestion
In particular, the FerretDB development cycle (fix-compile-run-dance) is faster when running on the host
```

b1ron (approved):
Made a small grammatical suggestion.

CONTRIBUTING.md
AlekSi: I think that changes the meaning: the suggested change says that the development cycle is faster when run on the host – but that's the only way we support it. The original sentence tries to say that the development cycle is faster when _FerretDB itself_ runs on a host.

b1ron: I understand.

b1ron (approved):

PR #672, opened 2023-11-28
Do not run mongo-tools test in parallel

b1ron:
This is just to rule out a possible race condition, see failure [here](https://github.com/FerretDB/dance/actions/runs/7019851965/job/19098277533?pr=634).


b1ron:
Still fails with `-p=1`.

* it always fails on the same collection
* it only fails with `mongoexport`
* it always fails after fetching the [first batch](https://github.com/FerretDB/dance/actions/runs/7023814419/job/19111541837#step:8:55)

PR #686, opened 2023-12-05
Bump deps

AlekSi:


b1ron (approved):
We are passing 3 more Go driver [tests](https://github.com/FerretDB/dance/actions/runs/7097333797/job/19317291268?pr=686#step:8:40). I can fix stats in a separate PR.

PR #770, opened 2024-02-13
Fix link to enmeshed submodule

b1ron:
Just fixes the link and updates the submodule.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,14 +2,13 @@
 	path = dumps/mongodb-sample-databases
 	url = https://github.com/mcampo2/mongodb-sample-databases.git
 	branch = master
-
 [submodule "dotnet-example"]
 	path = tests/dotnet-example
 	url = https://github.com/FerretDB/dotnet-example.git
 	branch = main
 [submodule "enmeshed-runtime"]
 	path = tests/enmeshed-runtime
-	url = https://github.com/nmshd/cns-runtime.git
+	url = https://github.com/nmshd/runtime.git
 	branch = main
 [submodule "java-example"]
 	path = tests/java-example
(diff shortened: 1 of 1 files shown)


henvic (approved):

.gitmodules
```diff
@@ -2,15 +2,10 @@
...
 	branch = master
-
 [submodule "dotnet-example"]
 	path = tests/dotnet-example
 	url = https://github.com/FerretDB/dotnet-example.git
 	branch = main
-[submodule "enmeshed-runtime"]
-	path = tests/enmeshed-runtime
-	url = https://github.com/nmshd/cns-runtime.git
-	branch = main
```
AlekSi, line 13: We need that line; otherwise, dependabot will not update that submodule.

There is also no reason to change the order and group of submodules. Just edit one line.

AlekSi (changes requested):

.gitmodules
b1ron: That was a result of doing `git rm` and then `git add submodule` etc., because I wasn't entirely sure if the submodule's code itself had changed.

Fixed.

chilagrow (dismissed):

.gitmodules
AlekSi: Well, the grouping change wasn't reverted. `dumps/` should be a separate group from `tests/` – as it was

.gitmodules
```diff
@@ -2,14 +2,13 @@
 	path = dumps/mongodb-sample-databases
 	url = https://github.com/mcampo2/mongodb-sample-databases.git
 	branch = master
-
```
AlekSi, line 5: Revert that change

AlekSi (approved):

rumyantseva (approved):

henvic (approved):

PR #784, opened 2024-02-20
Enable PostgreSQL authentication

### LINKED ISSUE #769: Improve tests for authentication (auth)
- [x] Re-add DB configurations for new auth.
- [x] Update documentation for the new dance.

b1ron:
Closes #769.


cmd/dance/main.go
```diff
@@ -131,6 +131,15 @@ func main() {
 
 		var runRes *config.TestResults
 
+		// this hack simplifies our configuration until we support SCRAM
+		if *dbF != "mongodb" {
+			// we only support PLAIN for now
```
b1ron, line 136: Factually incorrect. Will fix.

b1ron:
New [issue](https://github.com/FerretDB/dance/actions/runs/7992993540/job/21827846032?pr=784#step:9:1508), invalid salt length for the legacy shell.

docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
       # Always UTC+05:45. Set to catch timezone problems.
       - TZ=Asia/Kathmandu
       - FERRETDB_HANDLER=postgresql
-      - FERRETDB_POSTGRESQL_URL=postgres://user@postgres:5432/dance
+      - FERRETDB_POSTGRESQL_URL=postgres://username:password@postgres:5432/dance
```
b1ron, line 14: I changed the username to be consistent with FerretDB's Docker compose [file](https://github.com/FerretDB/FerretDB/blob/main/docker-compose.yml#L24).

AlekSi: That change means that authentication is effectively disabled for all tests that use MongoDB URI without credentials.

We need to remove them there, not update them.

tests/dotnet-example-auth-scram-sha-1.yml
```diff
@@ -4,7 +4,7 @@ dir: dotnet-example
 args:
   - dotnet
   - run
-  - mongodb://user:password@localhost:27017/?replicaSet=rs0&authMechanism=SCRAM-SHA-1
+  - mongodb://username:password@localhost:27017/?replicaSet=rs0&authMechanism=SCRAM-SHA-1
```
AlekSi, line 7: For example, that test would pass without username and password there

AlekSi (changes requested):

docker-compose.yml
AlekSi: In other words, that's not different from `trust`

b1ron: I see, when I do remove the credentials SCRAM authentication fails. Should we call `SetBypassBackendAuth()` before [here](https://github.com/FerretDB/FerretDB/blob/main/internal/handler/msg_saslstart.go#L81) to bypass backend authentication?

```
byron@byron-ThinkCentre-M910q:~/workdir/dance$ mongosh "mongodb://username:password@localhost:27017/?authMechanism=SCRAM-SHA-1"
Current Mongosh Log ID:	65d7722489a2ee64c03f12e8
Connecting to:		mongodb://<credentials>@localhost:27017/?authMechanism=SCRAM-SHA-1&directConnection=true&serverSelectionTimeoutMS=2000&appName=mongosh+1.10.6
MongoServerError: [msg_saslstart.go:184 handler.(*Handler).scramCredentialLookup] [collection.go:57 postgresql.(*collection).Query] [registry.go:279 metadata.(*Registry).DatabaseGetExisting] [registry.go:132 metadata.(*Registry).getPool] [pool.go:128 pool.(*Pool).Get] [opendb.go:95 pool.openDB] [opendb.go:110 pool.checkConnection] failed to connect to `host=postgres user=root database=dance`: failed SASL auth (FATAL: password authentication failed for user "root" (SQLSTATE 28P01))
```

AlekSi: We should ensure that the problem is not in the dance configuration, maybe test manually, then report the problem

b1ron:
SCRAM authentication fails due to FerretDB/FerretDB/issues/4115.

docker-compose.yml
b1ron: I tested it manually and it fails so I created an issue, see FerretDB/FerretDB/issues/4115.

.github/workflows/dance.yml
```diff
@@ -49,11 +49,10 @@ jobs:
           # - dbaas_core-2
           # - dbaas_core-3
           # - diff
-          - dotnet-example
           - dotnet-example-auth
           - dotnet-example-auth-scram-sha-1
           - dotnet-example-auth-scram-sha-256
-          - java-example
+          - enmeshed-runtime
```
AlekSi, line 55: It is checked in dance-trust.yml

PR #801, opened 2024-03-05
Update Python scripts to use the `venv` module

### LINKED ISSUE #768: Setup dependabot for example repositories
* https://github.com/FerretDB/dotnet-example
* https://github.com/FerretDB/java-example
* https://github.com/FerretDB/python-example

Those repos contain simple tests for language drivers, but there are no dependabot configurations that ensure that we test the latest versions. We should add them.

We also should add `requirements.txt` to the Python example as this repo does not use any dependency manager at all.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

b1ron:
Closes #768.


tests/python-example-auth-scram-sha-1.sh
```diff
@@ -2,6 +2,6 @@
 
 set -ex
 
-pip3 install pymongo
+pip3 install -r requirements.txt
```
AlekSi, line 5: Let's avoid installing those packages globally. The easiest way to do that is probably (still) virtualenv. Maybe there is something more modern and current

b1ron: Dependabot supports [Pipenv](https://pipenv.pypa.io/) which I think is the best candidate, because we would no longer need to use pip and virtualenv separately. This seems to solve our problem.

b1ron: And we also need to create a new virtualenv using the correct Python version installed on the system hence this slight [mess](https://github.com/FerretDB/dance/pull/801/files#diff-13a8e9762c839ef76971d9a70bce544a8957326a7e9f2074b9ed63163c64bef1R7).

tests/python-example-auth-scram-sha-1.sh
```diff
@@ -2,6 +2,11 @@
 
 set -ex
 
-pip3 install pymongo
+pip install --user pipenv
+
+pipenv --python $(python3 -c 'import sys; print(".".join(map(str, sys.version_info[:3])))')
+
+. $(pipenv --venv)/bin/activate
```
AlekSi, line 9: I'm not sure if `.` works correctly there. I think `python3` below would use a system Python, not the one from the virtualenv.

I think we could simplify that by using https://docs.python.org/3/library/venv.html

AlekSi (commented):
https://github.com/FerretDB/dance/pull/801#discussion_r1514869218

tests/python-example-auth-scram-sha-1.sh
b1ron: Yes, not sure how I ended up installing a third-party library. I am using the correct module and running the binaries from the virtual environment.

b1ron: :pray:

tests/python-example-auth-scram-sha-1.sh
```diff
@@ -2,6 +2,8 @@
 
 set -ex
 
-pip3 install pymongo
+python -m venv .
```
AlekSi, line 5:
```suggestion
python3 -m venv .
```

AlekSi (approved):

PR #857, opened 2024-04-30
Add nodejs-example

b1ron:
Until FerretDB/FerretDB/issues/4243 is done only MongoDB tests are valid on CI.


AlekSi (commented):
After #867

PR #867, opened 2024-05-06
Simplify CI

### LINKED ISSUE #866: Simplify CI setup
- [ ] Move all example repositories commands to scripts. For example, [this](https://github.com/FerretDB/dance/blob/5a2be70b5541cab221c28f02064a691bee296378/tests/dotnet-example-auth-scram-sha-1.yml#L5-L7) should be in the `dotnet-example-plain.sh` file.
- [ ] Merge SHA-1 and SHA-256 configurations scripts, name them like `java-example-scram` (drop the `auth` part, too).
- [ ] Update CI configuration.

b1ron:
Closes #866.


AlekSi (changes requested):
We should not remove tests for PLAIN authentication in this issue. It even mentions `dotnet-example-plain.sh` by name.

AlekSi (changes requested):
#866 says nothing about removing configurations without authentication. We should not do that.

We also should not remove restheart configuration.

Let's focus on doing what the issue says, not other unrelated things.

AlekSi (approved):

PR #1256, opened 2025-05-27
Allow flexibility to the format of measurements

### LINKED ISSUE #1202: Run MongoDB's mongo-core-test
https://github.com/FerretDB/mongo-core-test

No need to implement a custom runner or response parser yet.

Use `venv` like python-example does.

chilagrow:
Closes #1202.

The `mongodb-benchmarking` produces benchmark results that takes measurements every second while the benchmark is running. It looks like below. It doesn't fit well into the current measurement format of `map[string]float64`. This PR makes measurement format a bit more flexible.

```
t,count,mean,m1_rate,m5_rate,m15_rate,mean_rate
1748240899,13524,13522.216068,756.800000,756.800000,756.800000
1748240900,21505,10748.078321,756.800000,756.800000,756.800000
1748240901,27081,9027.363746,756.800000,756.800000,756.800000
1748240902,30000,8288.694528,756.800000,756.800000,756.800000
```

### DIFF
--- a/internal/pusher/pusher.go
+++ b/internal/pusher/pusher.go
@@ -131,7 +131,7 @@ func (c *Client) Push(ctx context.Context, config, database string, res map[stri
 
 	for t, tr := range res {
 		t = strings.ReplaceAll(t, ".", "_") // to make it compatible with FerretDB v1
-		passed = append(passed, bson.E{t, bson.D{{"m", tr.Measurements}}})
+		passed = append(passed, bson.E{t, bson.D{{"m", tr.Measurements.Values()}}})
 	}
 
 	doc := bson.D{
--- a/internal/runner/ycsb/ycsb.go
+++ b/internal/runner/ycsb/ycsb.go
@@ -49,6 +49,24 @@ type measurement struct {
 	Perc9999Us float64 `json:"99.99th(us),string"`
 }
 
+// Values implements [config.Measurements] interface.
+func (m *measurement) Values() any {
+	return map[string]float64{
+		"takes":    m.TakesS,
+		"count":    float64(m.Count),
+		"ops":      m.OPS,
+		"avg":      (time.Duration(m.AvgUs) * time.Microsecond).Seconds(),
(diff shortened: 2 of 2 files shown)


noisersup (dismissed):

AlekSi (commented):
> The mongodb-benchmarking produces benchmark results with measurements taken every second while the benchmark is running. It looks like below.
>
> ```
> t,count,mean,m1_rate,m5_rate,m15_rate,mean_rate
> 1748240899,13524,13522.216068,756.800000,756.800000,756.800000
> 1748240900,21505,10748.078321,756.800000,756.800000,756.800000
> 1748240901,27081,9027.363746,756.800000,756.800000,756.800000
> 1748240902,30000,8288.694528,756.800000,756.800000,756.800000
> ```

Okay…

> The format doesn't fit well into the current measurement format of map[string]float64 for each test.

But why? ycsb also produces many lines during its run. We just take the final line.

Currently, we don't take measurements during the run, only the final results. I don't think we should change that yet.

I'm not sure we need that change.

chilagrow:
> > The mongodb-benchmarking produces benchmark results with measurements taken every second while the benchmark is running. It looks like below.
> > ```
> > t,count,mean,m1_rate,m5_rate,m15_rate,mean_rate
> > 1748240899,13524,13522.216068,756.800000,756.800000,756.800000
> > 1748240900,21505,10748.078321,756.800000,756.800000,756.800000
> > 1748240901,27081,9027.363746,756.800000,756.800000,756.800000
> > 1748240902,30000,8288.694528,756.800000,756.800000,756.800000
> > ```
> 
> Okay…
> 
> > The format doesn't fit well into the current measurement format of map[string]float64 for each test.
> 
> But why? ycsb also produces many lines during its run. We just take the final line.
> 
> Currently, we don't take measurements during the run, only the final results. I don't think we should change that yet.
> 
> I'm not sure we need that change.

Ah I see, that case I can take the last measurement only.

PR #1258, opened 2025-05-27
Check config contains existing database.

chilagrow:
There is no `ferretdb-dev` it probably should be `ferretdb2-dev`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,8 +16,8 @@ linters-settings:
             desc: use `context` package instead
           - pkg: golang.org/x/exp/slices
             desc: use `slices` package instead
-          - pkg: maps
-            desc: use `golang.org/x/exp/maps` package instead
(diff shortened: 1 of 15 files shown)


.golangci.yml
```diff
@@ -16,8 +16,8 @@ linters-settings:
             desc: use `context` package instead
           - pkg: golang.org/x/exp/slices
             desc: use `slices` package instead
-          - pkg: maps
-            desc: use `golang.org/x/exp/maps` package instead
```
chilagrow, line 20: If `golang.org/x/exp/maps` is imported following linter complained.

```
internal/configload/configload.go:26:2: Import statement 'golang.org/x/exp/maps' can be replaced by 'maps' (exptostd)
        "golang.org/x/exp/maps"
```

AlekSi (approved):

